PR #37, opened 2015-02-18
Refactor msbuild project files

ermshiperete:
Refactor msbuild project files and enable NuGet package restore (especially for unit tests)

This will make it easier to get builds that work the same on TC, Jenkins, and the developer's machine.

These changes are similar to PR #184 in libpalaso.


chrisvire:
:+1: 

Goes along with https://github.com/sillsdev/libpalaso/pull/184

PR #38, opened 2015-03-04
Update Linux build script and instructions

jasonleenaylor:


jasonleenaylor:
Apparently environ-xulrunner is not necessary for Chorus to build and run tests from the command line as it had no affect in my testing.

jasonleenaylor:
I'm accepting 13 days of silence as approval.

PR #39, opened 2015-03-17
Update to Mercurial 3

jasonleenaylor:
This work was done to resolve two issues found in the field: 
* Sometimes Mercurial would perform a premerge on a file, declare it merged, and never call out to the specified merge tool.
* Sometimes Mercurial would hang during a merge of a file, not using any cycles, but never returning.

The first commit changes the way that we get ChorusMerge called by hg. This way allows us to tell hg to skip the premerge to fix the first issue.

The second commit includes all the changes required in our code in order to operate with Mercurial 3, and includes the updated zip of Mercurial for windows. Testing shows that with the same data Mercurial 3 does not hang when running the same scenario that hung 1.5.1

The third commit is the updated zip files for linux as well as some documentation for how to build those zip files.

CI builds of this branch can be seen here:
http://build.palaso.org/viewType.html?buildTypeId=Chorus_win32_master_NewHg
http://build.palaso.org/viewType.html?buildTypeId=Chorus_NewHgLinux

Integration of this version of Chorus with FLExBridge has been successfully smoke tested


MercurialExtensions/fixutf8/fixutf8.py
```diff
@@ -1,14 +1,15 @@
 # fixutf8.py - Make Mercurial compatible with non-utf8 locales
 #
 # Copyright 2009 Stefan Rusek
+# Copyright 2015 Jason Naylor
```
JohnThomson: Wouldn't it be more usual to use SIL as the copyright owner?

MercurialExtensions/fixutf8/fixutf8.py
```diff
@@ -185,41 +143,65 @@ def localize(h):
...
-	#print "extsetup start"
+	'''
+	The following function has nothing to do with utf8 and could potentially belong
+	in a separate extension. The old versions of mercurial allowed branch names that
+	were only integers. Starting in version 3 integer branch names were disallowed.
+	Chorus uses branch names to handle model upgrades. This change would force us to
+	simulate a model upgrade with Chorus clients that used only the model version number
+	as a branch name. Alternatively we can wrap the function and skip the integer name
+	test.
+	'''
```
JohnThomson: Presumably they were disallowed for a reason. Are you pretty sure that just bypassing the test will make everything fine? No commands we use that will recognize numbers as something other than branch names, for example?

I suppose we can't much help it since our current products have created all kinds of repositories which DO have numbers as branch names. But it wouldn't be very hard to enhance Chorus so that future version branches were named v1.2 etc.

MercurialExtensions/fixutf8/fixutf8.py
```diff
@@ -185,41 +143,65 @@ def localize(h):
...
+def checklabelwrapper(orig, repo, lbl, kind):
+	try:
+		int(lbl)
+		pass
+	except ValueError:
+		return orig(repo, lbl, kind)
+
+
+def extsetup():
 	if sys.platform != 'win32':
```
JohnThomson: I suppose this is also used for win64? Sigh.

MercurialExtensions/fixutf8/fixutf8.py
regnrand: They are only accidentally integers. For instance, had we enabled branching for model versions in Lift data, it would have been: 0.13, not something that looks like an integer, as in FLEx's model version number/branch name.

MercurialExtensions/fixutf8/win32helper.py
```diff
@@ -113,6 +242,212 @@ def getargs():
...
+				s = data.value[0:bytesRead.value]
+				# print("\nv: %s" % repr(s), file=sys.stderr)
+				func(fromunicode(s))
+
+		def run(self, stdout_func=None, stderr_func=None):
+			"""Runs the process, using the provided functions for I/O.
+			The functions stdout_func and stderr_func are called whenever
+			something is printed to stdout or stderr, respectively.
+			These functions are called from different threads (but not
+			concurrently, because of the GIL).
```
JohnThomson: GIL doesn't communicate to me...maybe it would to someone who knows Python better (not hard) or who knows the Hg source better (also not at all hard)

lib/common/UpdatingMercurial.md
```diff
@@ -0,0 +1,16 @@
+To update the mercurial zip files for the linux distributions you should perform the following steps or some reasonable facsimile:
```
JohnThomson: Is there anything similar to do for Windows? This file doesn't seem to be in a Linux-specific folder.

src/Installer/GeneratedMercurialExtensions.wxs
```diff
@@ -1,71 +1,91 @@
 ﻿<?xml version="1.0" encoding="utf-8"?>
 <Wix xmlns="http://schemas.microsoft.com/wix/2006/wi">
-	<Fragment>
```
JohnThomson: I'm surprised this file is checked in, if it is generated as its name implies

regnrand: It is a generated file, but not as a normal part of any build. It is rebuilt by calling a special build target, and then only when some files are known to have changed.

src/LibChorus/VcsDrivers/Mercurial/HgRepository.cs
```diff
@@ -218,32 +214,49 @@ private static void CheckMercurialIni()
...
+			var mergetoolname = "chorusmerge";
+			var doc = GetMercurialConfigForRepository();
+			var chorusMergeLoc = SurroundWithQuotes(ExecutionEnvironment.ChorusMergeFilePath());
+			var uiSection = doc.Sections.GetOrCreate("ui");
+
+			if(uiSection.GetValue("merge") != mergetoolname)
+			{
+				uiSection.Set("merge", mergetoolname);
+				var mergeToolsSection = doc.Sections.GetOrCreate("merge-tools");
+				mergeToolsSection.Set(String.Format("{0}.premerge", mergetoolname), "False");
```
JohnThomson: If I'm understanding at all, this is the core of how we fixed a problem that required weeks of effort. I think it deserves a comment linking it to some of the issues and explaining why premerge is a disaster for Chorus.

src/LibChorus/VcsDrivers/Mercurial/HgRepository.cs
```diff
@@ -600,13 +601,35 @@ private ExecutionResult Execute(bool failureIsOk, bool noChangeIsOk, int seconds
...
 		{
-			_progress.WriteMessageWithColor(@"Blue", "Attempting to recover from failed merge.");
-			var result = Execute(false, SecondsBeforeTimeoutOnMergeOperation, "resolve", "--all");
-			if (!String.IsNullOrEmpty(result.StandardError))
+			var heads = GetHeads();
+			if(heads.Count() > 2)
+			{
+				throw new ApplicationException(String.Format("Unable to recover from failed merge: [Too many heads in the repository]"));
+			}
+			MergeSituation.PushRevisionsToEnvironmentVariables(heads[0].UserId, heads[0].Number.Hash,
```
JohnThomson: A comment might help here. I have no idea why we want to push things into temporary environment variables as part of recovering from a failed merge. It also doesn't feel related to the problem this PR is about.

src/LibChorus/VcsDrivers/Mercurial/HgRepository.cs
```diff
@@ -1475,10 +1471,28 @@ private static void SetExtensions(IniDocument doc, IEnumerable<KeyValuePair<stri
 			}
 		}
 
-		internal static bool CheckExtensions(IniDocument doc, IEnumerable<KeyValuePair<string, string>> extensionDeclarations)
+		/// <summary/>
+		/// <returns>
+		/// Returns true if the extensions in the repo contain only the extensions declared in the given map.
+		///</returns>
+		internal static bool CheckExtensions(IniDocument doc, Dictionary<string, string> extensionDeclarations)
```
JohnThomson: I suppose this extensions file we're messing with is private to Chorus? We wouldn't want to mess up the set of extensions the user is using to manage something else.
How important is it that no other extensions are active? Hg will, I believe, look for a global settings file which might also configure extensions. Or do we do something to suppress that? I'm wondering whether we need to do something here to explicitly DISable any known extensions that cause problems.

src/LibChorus/VcsDrivers/Mercurial/HgRepository.cs
```diff
@@ -1475,10 +1471,28 @@ private static void SetExtensions(IniDocument doc, IEnumerable<KeyValuePair<stri
...
 		{
-			var section = doc.Sections.GetOrCreate("extensions");
-			return extensionDeclarations.All(pair => section.GetValue(pair.Key) == pair.Value);
```
JohnThomson: I rather like doing it in one line with All (would be two lines now, of course).

src/LibChorus/merge/xml/generic/MergeChildrenMethod.cs
```diff
@@ -57,6 +57,16 @@ public void Run()
...
 			var parentStrategy = _merger.MergeStrategies.GetElementStrategy(parentNode);
 			var parentOrder = parentStrategy == null ? ChildOrder.AskChildren : parentStrategy.ChildOrderPolicy.OrderSignificance(parentNode);
+			// Determine if child order is important, we shouldn't create reordering conflicts if the order of children is unimportant
+			var childOrderMatters = parentOrder == ChildOrder.Significant;
+			if(parentOrder == ChildOrder.AskChildren && parentNode.HasChildNodes)
+			{
+				var childStrategy = _merger.MergeStrategies.GetElementStrategy(parentNode.FirstChild);
+				if(childStrategy != null && childStrategy.OrderIsRelevant)
+				{
+					childOrderMatters = true;
```
JohnThomson: Surprising that it depends on just the first child...is that how it is documented? As opposed to, say, we need to consider order if any of the children says that it matters?

src/LibChorus/VcsDrivers/Mercurial/HgRepository.cs
regnrand: Yes, the ini file is only for the specific Hg repo. Hg has a global one, but we don't mess with it.

src/LibChorusTests/VcsDrivers/Mercurial/HgSettingsTests.cs
```diff
@@ -354,6 +354,30 @@ public void EnsureTheseExtensionAreEnabled_noExistingExtensions_AddsThem()
 		}
 
 		[Test]
+		public void EnsureTheseExtensionAreEnabled_ExtraExtension_RemovesThem()
```
JohnThomson: Suggest ExtraExtensions to match 'Them'...made it hard to parse.

src/LibChorusTests/VcsDrivers/Mercurial/HgSettingsTests.cs
```diff
@@ -354,6 +354,30 @@ public void EnsureTheseExtensionAreEnabled_noExistingExtensions_AddsThem()
...
+				var extensions = new Dictionary<string, string>();
+				extensions.Add("a", "");
+				extensions.Add("b", "");
+				repository.EnsureTheseExtensionsAndFormatSet(extensions);
+
+				Assert.AreEqual(2, repository.GetEnabledExtension().Count());
+				Assert.AreEqual("a", repository.GetEnabledExtension().ToArray()[0]);
+				Assert.AreEqual("b", repository.GetEnabledExtension().ToArray()[1]);
+			}
+		}
```
JohnThomson: I don't see anything in this test that requires EnsureTheseExtensions... to remove any extra ones

src/LibChorusTests/VcsDrivers/Mercurial/HgSettingsTests.cs
```diff
@@ -354,6 +354,30 @@ public void EnsureTheseExtensionAreEnabled_noExistingExtensions_AddsThem()
...
+
+		[Test]
 		public void EnsureTheseExtensionAreEnabled_someOthersEnabledAlready_StayEnabled()
```
JohnThomson: It seems the test no longer proves what it's name promises, because the SUT now intentionall does NOT keep pre-existing extensions enabled. Needs renaming at least.

src/LibChorusTests/VcsDrivers/Mercurial/HgVersionBranchingTests.cs
```diff
@@ -127,5 +127,45 @@ public void DoesNewBranchExist_No()
...
+				repoWithFiles.SyncWithOptions(new SyncOptions
+				{
+					DoPullFromOthers = false,
+					CheckinDescription = "version number branch",
+					DoSendToOthers = false
+				});
+
+				const string myVersion = ""; // Hg default branch name
+
+				// SUT
```
JohnThomson: It's not obvious that this is the SUT. Isn't the point of this test mainly that the SyncWithOptions is actually able to make a new branch with an integer name?

src/LibChorusTests/VcsDrivers/Mercurial/HgVersionBranchingTests.cs
```diff
@@ -127,5 +127,45 @@ public void DoesNewBranchExist_No()
...
+				string revNum;
+				bool result = branchingHelper.IsLatestBranchDifferent(myVersion, out revNum);
+
+				// Verification
+				Assert.IsTrue(result, "The only branch should be default.");
+				var revision = repoWithFiles.Repository.GetRevision(revNum);
+				Assert.AreEqual(integerBranchName, revision.Branch, "Wrong branch name in new branch.");
```
JohnThomson: Hmmm, I guess not, since we're verifying that we did NOT get a new branch with the integer name. I'm thoroughly confused after several minutes studying this method, so I think something needs improving...naming, comments,...partly because not seeing this in an IDE I have to try to remember/guess what on earth a SnychronizerAdjunct and its constructor argument are.

JohnThomson:
Nice piece of very difficult work overall. My points are mostly quibbles.

src/LibChorus/VcsDrivers/Mercurial/HgRepository.cs
JohnThomson: Right, but Jason is going to some trouble to make sure there are no extensions except the ones we want. I don't know why, but if it's crucial that something NOT be active, we might need to worry about the global settings activating it.

src/LibChorus/merge/xml/generic/MergeChildrenMethod.cs
jasonleenaylor: As I understand it, the generic xml does not support having children with different strategies. There are other places in this area that assume only the first child under a parent matters.

MercurialExtensions/fixutf8/fixutf8.py
jasonleenaylor: As far as I can tell they just decided that it was a bad idea. They do not force old repositories to change number only branch names as that would break their backward compatibility. All of their code must still work with them. We could certainly avoid use of number only branch names in the future just in case they do something more drastic in Mercurial 4, but that change would need to be made in the clients.

MercurialExtensions/fixutf8/fixutf8.py
jasonleenaylor: Apparently. (shrug)

MercurialExtensions/fixutf8/win32helper.py
jasonleenaylor: I had to look it up myself. It's 'Global Interpreter Lock'. I guess I could change the comment to be more verbose, but anyone who knows what the 'Global Interpreter Lock' in python is also knows the abbreviation.

src/LibChorus/VcsDrivers/Mercurial/HgRepository.cs
jasonleenaylor: True. Done.

src/LibChorus/VcsDrivers/Mercurial/HgRepository.cs
jasonleenaylor: Well, the short answer is that the test for this situation failed in Mercurial 3 so I was forced to evaluate it and I discovered that it probably didn't ever work at all. Pushing to the environment is required for our merge tool to be used during the recovery attempt. Our merge tool was not actually used in the recovery attempt before, so whenever this happened historically it almost certainly corrupted data.

src/LibChorus/VcsDrivers/Mercurial/HgRepository.cs
jasonleenaylor: I believe that the mercurial.ini in the working directory will override the global one. We formerly used a deprecated extension that Mercurial 3 will no longer let us use. That is why it is important that we remove extensions that we don't currently expect. Even if the global settings somehow came into play, they wouldn't actually have settings that are critical to remove or mercurial would be broken system wide.

src/LibChorus/VcsDrivers/Mercurial/HgRepository.cs
jasonleenaylor: The Linq version seems reasonably legible. Done.

src/LibChorusTests/VcsDrivers/Mercurial/HgSettingsTests.cs
jasonleenaylor: Done

src/LibChorusTests/VcsDrivers/Mercurial/HgSettingsTests.cs
jasonleenaylor: The .hgrc was created with 'a', and 'x' while the dictionary has 'a' and 'b'. The count on the result combined with verifying 'a' and 'b' in the result proves 'x' was removed.
I've taken the opportunity to add appropriate failure messages on the asserts.

src/LibChorusTests/VcsDrivers/Mercurial/HgSettingsTests.cs
jasonleenaylor: Done.

src/LibChorusTests/VcsDrivers/Mercurial/HgVersionBranchingTests.cs
jasonleenaylor: Hmm, yes, this test is confusing. It had the misfortune of being the last test I wrote, and of doing its job after a quick copy paste and tweak. I'll give it a bit of attention.

src/LibChorus/VcsDrivers/Mercurial/HgRepository.cs
JohnThomson: Seems it would be useful to say at least "Pushing to the environment is required for our merge tool to be used during the recovery attempt." in a comment.

JohnThomson: I'm not at all clear what this method is trying to do. How did the merge fail, and what do we think our tool might be able to do about it? Isn't our tool always used to do the merge? If so, and it already failed, why would we think it will now be able to repair things? A comment on the whole method might help.

src/LibChorus/VcsDrivers/Mercurial/HgRepository.cs
JohnThomson: My impression is that local settings override but do not completely replace global ones. That might be worth checking. In any case it would be worth commenting that a previous version might have left harmful settings in the repo's config file.

src/LibChorus/merge/xml/generic/MergeChildrenMethod.cs
JohnThomson: I see. So the parent can have more than one kind of child, but they'd better all be the same for any given parent. Fair enough.

MercurialExtensions/fixutf8/win32helper.py
JohnThomson: Yes, the longer name would not help. What I think I would most want to know is whether the reason is something clever we do somewhere (that a change to our code might break) or something inherent in Python. Something like "because python keeps this sort of thing sequential using a global lock (GIL)". You may know enough to be more specific than "this sort of thing" without being too verbose.

jasonleenaylor: I did my best. :)

src/LibChorus/VcsDrivers/Mercurial/HgRepository.cs
jasonleenaylor: Comment added in method remarks.

PR #40, opened 2015-03-25
Sign ChorusHubInstaller

chrisvire:
It is confusing to users to download/install ChorusHubInstaller.msi
without it being signed.  Create a new "SignChorusHub" target that
will be run by the Team City build
"ChorusHub master Windows Service Installer-sans publish"

I have tested this change by pointing the above mentioned build
to this change in my fork of this repo.  See build 2.4.14 of that build.

### DIFF
--- a/build/Chorus.proj
+++ b/build/Chorus.proj
@@ -185,6 +185,11 @@
 
 	</Target>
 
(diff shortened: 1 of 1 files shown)


jasonleenaylor:
Looks like something we want to do.

PR #41, opened 2015-03-30
Remove unnecessary ini modification from merge module

jasonleenaylor:
* Delete the section from the merge module source which is
  unnecessary after recent Chorus modifications and which is
  also invalid (due to cross component linking) in new Wix versions.


regnrand:
LGTM

PR #43, opened 2015-03-30
Repent and fix merge module

jasonleenaylor:
After making several mistakes I have learned the reasons for all the things I removed this Monday.
This PR reverts two changes and makes a different fix for the installer problem.


regnrand:
LGTM, but then so did the last attempts to fix it. :-)

PR #47, opened 2015-04-24
Handle whole file deletion vs changed robustly

jasonleenaylor:
* Detect and respond to Hg information about file deletion vs change conflicts which could block a merge.
* Create new conflict for this situation.
* Rename ProcessOutputReader to HgProcessOutputReader  to accurately reflect its purpose


regnrand:
LGTM.

PR #48, opened 2015-04-29
Update to ICU 54

ddaspit:
* update buildupdate scripts
* update LibChorus.csproj to copy new ICU DLLs
* remove old Palaso DLLs and compiled python files


ddaspit:
I already started merging in the ICU 54 changes to Palaso and ICU.NET before I realized that Chorus would also need to be updated. This is a minor change and we've already got buy-in for the change from the Palaso pull request, so I am just going to go ahead and merge this in myself.

PR #49, opened 2015-04-29
Address the root cause of backward compatibility issues for USB

jasonleenaylor:
* Stop updating the hgrc files on an Identifier check
* Don't set any extensions after cloning to USB
* Adjust the method naming so that it is clear when we are cloning
  to USB instead of from
* Fix utf8 extension to allow us to use the --config option


MercurialExtensions/fixutf8/win32helper.py
```diff
@@ -233,15 +233,25 @@ def consolehascp():
...
+		Skips any --config and following arguments since this is used in a method where
+		those arguments are already removed
 		'''
 		c = INT(0)
 		pargv = CommandLineToArgv(GetCommandLine(), byref(c))
-		return [fromunicode(pargv[i]) for i in xrange(1, c.value)]
+		cleanArguments = []
+		iterator = iter(xrange(1, c.value))
+		for i in iterator:
+			if pargv[i] != "--config":
```
JohnThomson: I don't know python much, but this looks as if it will skip --config, but not any following arguments. There's nothing like a break, it only checks the current one to see whether it IS --config.

src/LibChorus/VcsDrivers/Mercurial/HgHighLevel.cs
```diff
@@ -6,9 +6,19 @@ namespace Chorus.VcsDrivers.Mercurial
...
+
+		public static string MakeCloneFromLocalToUsb(string sourcePath, string targetDirectory, IProgress progress)
+		{
+			return MakeCloneFromLocalToLocal(sourcePath, targetDirectory, false, progress);
+		}
+
 		//TODO: get rid of this, or somehow combine it with the other Clone() options out there
 		/// <returns>path to clone, or empty if it failed</returns>
-		public static string MakeCloneFromLocalToLocal(string sourcePath, string targetDirectory, bool alsoDoCheckout, IProgress progress)
+		private static string MakeCloneFromLocalToLocal(string sourcePath, string targetDirectory, bool alsoDoCheckout, IProgress progress)
```
JohnThomson: This variable now affects another behavior as well, might be good to rename it, maybe something like fromUsb?

src/LibChorusTests/VcsDrivers/Mercurial/RepositoryTests.cs
```diff
@@ -140,18 +140,22 @@ public void RepositoryRecoversFromIncompleteMerge()
...
-		public void CloneToUsbWithoutUpdate_CreatesExtensionSectionInHgrc()
+		public void CloneToUsbWithoutUpdateFollowedByIdentifierDoesNotAffectHgrc()
 		{
 			using(var repo = new RepositorySetup("source"))
 			using(var f = new TemporaryFolder("clonetest"))
 			{
 				// The MakeCloneFromLocalToLocal with false on alsoDoCheckout is the core of the usb clone operation.
-				// We need to make sure that the extensions are used in the clone
-				HgHighLevel.MakeCloneFromLocalToLocal(repo.ProjectFolder.Path, f.Path, false, new NullProgress());
+				// We need to make sure that this clone is bare of extensions, and remains so after the identifier is checked.
```
JohnThomson: This sounds as if you will force the clone not to have any extensions at all, rather than just being careful not to change anything, which sounds more reasonable for an identity check. Is that really what you mean?

src/LibChorus/VcsDrivers/Mercurial/HgHighLevel.cs
jasonleenaylor: good point. I'll rename it.

MercurialExtensions/fixutf8/win32helper.py
jasonleenaylor: The iterator.next() in the else block accomplishes skipping the following argument. We don't add config and we skip the following. It probably deserves a comment.

JohnThomson: I suppose Python lets you get away with that...modifying an iterator within the loop using it strikes me as dangerous.
The other thing that threw me off is that the comment above says "Skips any --config and following arguments" which I read as meaning EVERYTHING after --config, not just the next one.

jasonleenaylor: I clarified the comment that confused you.

PR #55, opened 2015-06-04
Fix CHR-32: Fix the fixutf8.py mercurial extension to work with CP1252

cambell-prince:
Fix CHR-32: Fix the fixutf8.py mercurial extension to work with CP1252

- Restores code in fixutf8.py to permit working with CP1252 strings
   in addition to unicode.

- Added a test that works to catch this.


src/LibChorus.TestUtilities/RepositorySetup.cs
```diff
@@ -219,9 +217,11 @@ public static void MakeRepositoryForTest(string newRepositoryPath, string userId
 			hg.SetUserNameInIni(userId,  progress);
 		}
 
-		public static string ProjectName
+		public string ProjectName { get; private set; }
+
+		public static string ProjectNameForTest
 		{
-			get { return "ไก่ projéct"; } //nb: important that it have a space, as this helps catch failure to enclose in quotes
```
hatton: With this change, several tests that were essentially testing the unicode path will now instead be testing the CP1252 path. If you're happy with that, I am too, just pointing it out.

cambell-prince: Yes, I decided that seeing as this was the option that broke, I'd switch to having CP1252 being the greater test coverage.  I've also added tests in the utf8tests to cover either option.  Previously these tests were defective; 'CreateOrUseExisting' was not suffcient to test, and RecoverFromInteruptedTransaction which showed the failure in client apps was not tested anywhere else in Chorus.  So, now we should have at least one test that works for each of unicode and cp1252.

jasonleenaylor:
LGTM, thanks for mopping up.

PR #61, opened 2015-08-11
Increase Chorus version number to 2.5

jasonleenaylor:
* This time change the version in the build file that is
  actually used to produce the TeamCity windows artifacts


jasonleenaylor:
Randy says: If the build is happy, go ahead and merge it in.

PR #67, opened 2015-09-02
Don't Crash Showing Diagnostic Information

papeh:
`hg showconfig` can return output with braces, which crashes `String.Format`.
Escape braces.


papeh:
Cambell: At your suggestion, I wrote a test, and discovered that most tests use a different progress-logger that handles this crash better.  I will refactor my change accordingly.

PR #71, opened 2015-09-17
Do Not Fail the Pull if One Chunk Fails

papeh:
If one chunk in a pull fails, retry; don't mark the whole pull as a failure.

### DIFF
--- a/src/LibChorus/VcsDrivers/Mercurial/HgResumeTransport.cs
+++ b/src/LibChorus/VcsDrivers/Mercurial/HgResumeTransport.cs
@@ -710,14 +710,10 @@ namespace Chorus.VcsDrivers.Mercurial
 				}
 				if (response.Status == PullStatus.Fail)
 				{
-					// Not sure that we need to abort the attempts just because .Net web request says so.  See Push also. CP 2012-06
-					errorMessage = "Pull data chunk failed";
-					_progress.WriteError(errorMessage);
+					// No need to abort the attempts just because .Net web request says so. See Push also. CP 2012-06
+					_progress.WriteWarning("Pull data chunk failed");
 					retryLoop = true;
 					req.StartOfWindow = bundleHelper.StartOfWindow;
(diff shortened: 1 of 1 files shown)


src/LibChorus/VcsDrivers/Mercurial/HgResumeTransport.cs
```diff
@@ -710,14 +710,10 @@ public bool Pull(string[] baseRevisions)
 				}
 				if (response.Status == PullStatus.Fail)
 				{
-					// Not sure that we need to abort the attempts just because .Net web request says so.  See Push also. CP 2012-06
-					errorMessage = "Pull data chunk failed";
-					_progress.WriteError(errorMessage);
```
papeh: WriteError sets a flag that we later check to determine whether the get was successful. Since this doesn't phase us; use Warning instead.

cambell-prince: I'm interested in the cases where this is true, where do I find out more info?  TCP is a reliable transport, so the server *choosing* to send this response is interesting.

papeh: From the log where we saw this:

    Warning: Pull data chunk failed: Error: ConnectFailure (interrupted)
    Error: Pull data chunk failed

`PullStatus.Fail` is the default return of `PullOneChunk`; it is returned for some instances of `HttmStatusCode.BadRequest`, for an unrecognized server response, or if a `WebException` is thrown.  That's all I know.

cambell-prince: Seems like WriteError has a whole heap of power if that can cause the pull to fail. :-/  Looks more like a WriteErrorAndIndicateTotalFailureOfWholeProcess to me.  Would be better if those were to be separated at some point I think.

PR #76, opened 2015-11-09
Update dependencies for Linux

ermshiperete:
Download NDesk instead of having it in the repo.

[Review on Reviewable](https://reviewable.io/reviews/sillsdev/chorus/76)


jasonleenaylor:
Are you sure that this is the right NDesk? When we updated the dll the last time there was no continuous build available for Tom's fork of NDesk which had the bug fixes we needed. I'm just making sure that we don't have a very nasty to track down bug coming back on us.


---







Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/sillsdev/chorus/76)

jasonleenaylor:
Reviewed 5 of 5 files at r1.

---



Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/sillsdev/chorus/76)

ermshiperete:
Well, this gets the one from libpalaso. I don't know if that is the right one - I'd hope so since it gets used in libpalaso - I have no idea where I would find the "right" one otherwise.


---







Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/sillsdev/chorus/76)

PR #80, opened 2015-11-20
Consistently perform update when cloning or pushing to a USB drive

ddaspit:
When we push to a USB drive, we perform an update, populating the working directory. This behavior was inconsistent with when we clone to a USB drive. This inconsistency is causing problems with WeSay. I updated the code that clones to a USB drive so that it also performs an update. The alternative is to make both the push and clone to a USB drive not perform updates. From comments in Synchronizer.SendToOneOther(), performing an update on a USB is desirable, so I felt it was better to make the clone also perform an update.

### DIFF
--- a/src/LibChorus/VcsDrivers/Mercurial/HgHighLevel.cs
+++ b/src/LibChorus/VcsDrivers/Mercurial/HgHighLevel.cs
@@ -1,5 +1,4 @@
-﻿using System.IO;
-using SIL.Code;
+﻿using SIL.Code;
 using SIL.Progress;
 
 namespace Chorus.VcsDrivers.Mercurial
@@ -36,13 +35,10 @@ namespace Chorus.VcsDrivers.Mercurial
 			{
 				// Make a backward compatibile clone if cloning to USB (http://mercurial.selenic.com/wiki/UpgradingMercurial) 
 				targetDirectory = local.CloneLocalWithoutUpdate(targetDirectory, cloningFromUsb ? null : "--config format.dotencode=false --pull");
-				File.WriteAllText(Path.Combine(targetDirectory, "~~Folder has an invisible repository.txt"), "In this folder, there is a (possibly hidden) folder named '.hg' that contains the actual data of this Chorus repository. Depending on your Operating System settings, that leading '.' might make the folder invisible to you. But Chorus clients (WeSay, FLEx, OneStory, etc.) can see it and can use this folder to perform Send/Receive operations.");
 
-				if (cloningFromUsb)
-				{
-					var clone = new HgRepository(targetDirectory, progress);
-					clone.Update();
-				}
+				var clone = new HgRepository(targetDirectory, cloningFromUsb, progress);
+				clone.Update();
+
 				return targetDirectory;
 			}
 		}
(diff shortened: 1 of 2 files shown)


jasonleenaylor:
While I wasn't involved in the design I'm pretty sure the plan was to never perform an update on the USB stick. That's why the '~~Folder has an invisible repository.txt' file and code to write it even exists. I'm surprised that this could cause problems for WeSay. How does the update situation the USB stick come into play at all?

regnrand:
At the time, it was intentional to never do an update on the USB drive.

Randy

ddaspit:
@jasonleenaylor, I certainly understand the logic, but it doesn't seem to be working that way right now. The first time you do a send/receive to a USB drive, it performs a clone without an update and inserts the text file. The next time you do a send/receive to the USB drive, it performs a push and then an update, which populates the working directory with repo files. This seems like inconsistent behavior, where the push is negating the purposeful behavior of the clone. If you look at the code comment in the Synchronizer.SendToOneOther() method, it seems that performing an update on a USB drive is the desired behavior, but if you look at the clone code in HgHighLevel.MakeCloneFromLocalToLocal(), it seems that the opposite behavior is desired. We should probably pick one or the other. Do we want to perform updates on USB drives or not? I can certainly change it so that Synchronizer.SendToOneOther() does not perform an update after a push on a USB drive if that is the better choice.

regnrand:
Damien: "Do we want to perform updates on USB drives or not?"

Randy: John Hatton decided way back that it would not do the updates, and I fixed Chorus to not do them. It seems that someone then changed part of the code to do the update. I don't know if John changed his mind on the update, or if the one update change came in without approval. John has changed his mind on things from time to time, but I don't know if this is one of those cases, or not. I do know that doing the updates has caused problems with users in the past: 1) edits on the USB getting 'lost', since they were not committed, and 2) user concerns about repo bloat on the USB drive, which concern was mitigated when the user understood the update wasn't needed or part of the repo. So, without a change of mind from John, the update ought not be done.

Damien: "This inconsistency is causing problems with WeSay."
Randy: I don't understand how it can cause problems for any client. The Chorus clients don't even know where the S/R was done. Can you expand on the nature of the problem you are seeing?

regnrand:
The correct code is to *not* update the USB drive. So, this PR should be abandoned and the Chorus code that does the update should be fixed to not do it.

There is also no reason to update the Chorus Hub repos, if that is also being done.

regnrand:
This is the preferred fix for the other end (Synchronizer.SendToOneOther()):

// The update is only done for tests, since only tests now use "DirectoryRepositorySource".
if ((address is DirectoryRepositorySource && ((DirectoryRepositorySource)address).LooksLikeLocalDirectory))
{
	// passes false to avoid updating the hgrc on a usb send to preserve backward compatibility
	// passes false to avoid updating the hgrc to preserve backward compatibility
	var otherRepo = new HgRepository(resolvedUri, false, _progress);
	otherRepo.Update();
}

UsbRepositorySourceTests, then will have one test that will fails, so it needs this fix:

1. Orig test code:
//did it update too (which we should do with usb, unless we switch to leave them as "bare" .hgs)?
contents = File.ReadAllText(Path.Combine(usbDirectory, "foo.txt"));
Assert.AreEqual("version two", contents);

2. Revised test code:
// Should *not* update USB.
// NB: 'foo.txt' ought not be written to the faux USB folder in the first place,
// so it shouldn't even be there to read, but....
Assert.AreEqual("version one", contents);

ddaspit:
I am going to abandon this pull request and submit a new one that no longer performs an update when pushing to a USB drive.

PR #81, opened 2015-11-24
Ensure that an existing target repo on a USB drive is "bare" after sync

ddaspit:
Chorus purposefully clones a "bare" repo to a USB drive. Unfortunately, when pushing to an existing USB repo, Chorus performs an update, causing the repo to no longer be "bare". This pull request corrects this inconsistency by no longer performing the update to a USB repo when pushing.

[Review on Reviewable](https://reviewable.io/reviews/sillsdev/chorus/81)

### DIFF
--- a/src/LibChorus/sync/Synchronizer.cs
+++ b/src/LibChorus/sync/Synchronizer.cs
@@ -291,18 +291,12 @@ namespace Chorus.sync
 						repo.Push(address, resolvedUri);
 					}
 
-					// For LAN, it could be safe and desirable to do an update... for now we assume it is.
-					// For me (RandyR) including the shared network folder
-					// failed to do the update and killed the process, which left a 'wlock' file
-					// in the shared folder's '.hg' folder. No more S/Rs could then be done,
-					// because the repo was locked.
-					// For now, at least, it is not a requirement to do the update on the shared folder.
-					// JDH Oct 2010: added this back in if it doesn't look like a shared folder
 					// For USB, we do not wish to do an update, since it can cause problems if the working
-					// files are available to the user
+					// files are available to the user.
+					// The update is only done for tests, since only tests now use "DirectoryRepositorySource".
 					if (address is DirectoryRepositorySource && ((DirectoryRepositorySource) address).LooksLikeLocalDirectory)
 					{
-						// passes false to avoid updating the hgrc on a usb send to preserve backward compatibility
+						// passes false to avoid updating the hgrc on a send to preserve backward compatibility
 						var otherRepo = new HgRepository(resolvedUri, false, _progress);
 						otherRepo.Update();
 					}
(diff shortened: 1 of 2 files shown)


regnrand:
Reviewed 2 of 2 files at r1.
Review status: all files reviewed at latest revision, 4 unresolved discussions, some commit checks failed.

---

_[src/LibChorus/sync/Synchronizer.cs, line 294 \[r1\]](https://reviewable.io:443/reviews/sillsdev/chorus/81#-K3tzdcH2tBNCNeIYBaA:-K3tzdcH2tBNCNeIYBaB:869689785) ([raw file](https://github.com/sillsdev/chorus/blob/0325b762589799ac6661923067d5b1115e933bfa/src/LibChorus/sync/Synchronizer.cs#L294)):_
The references to LAN can be removed, since that option was repalced with ChorusHub, which does not DirectoryRepositorySource, but its own subclass of RepositorySource. DirectoryRepositorySource is now only used with testing to emulate a USB drive and/or ChorusHub.

---

_[src/LibChorus/sync/Synchronizer.cs, line 302 \[r1\]](https://reviewable.io:443/reviews/sillsdev/chorus/81#-K3u02Djn0XOmXyatgwo:-K3u02Djn0XOmXyatgwp:909988294) ([raw file](https://github.com/sillsdev/chorus/blob/0325b762589799ac6661923067d5b1115e933bfa/src/LibChorus/sync/Synchronizer.cs#L302)):_
Please add this to the end of the main comment to explain the use of DirectoryRepositorySource here:

// The update is only done for tests, since only tests now use "DirectoryRepositorySource".

In a better world, "DirectoryRepositorySource" would be removed altogether, and some other approach to testing would be used, say that used an interface for the various RepositorySource implementations.

---

_[src/LibChorus/sync/Synchronizer.cs, line 305 \[r1\]](https://reviewable.io:443/reviews/sillsdev/chorus/81#-K3u-F4SUq-8Yt6vwXJx:-K3u-F4SUq-8Yt6vwXJy:-775671082) ([raw file](https://github.com/sillsdev/chorus/blob/0325b762589799ac6661923067d5b1115e933bfa/src/LibChorus/sync/Synchronizer.cs#L305)):_
This comment should also be changed to something like: "passes false to avoid updating the hgrc on a send to preserve backward compatibility"

---

_[src/LibChorusTests/sync/UsbRepositorySourceTests.cs, line 108 \[r1\]](https://reviewable.io:443/reviews/sillsdev/chorus/81#-K3tzUnJHSfpiwKZ9v_g:-K3tzUnJHSfpiwKZ9v_h:-1849740105) ([raw file](https://github.com/sillsdev/chorus/blob/0325b762589799ac6661923067d5b1115e933bfa/src/LibChorusTests/sync/UsbRepositorySourceTests.cs#L108)):_
Good fix.

---


_Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/sillsdev/chorus/81)_

jasonleenaylor:
Reviewed 2 of 2 files at r1.
Review status: all files reviewed at latest revision, 4 unresolved discussions, some commit checks failed.

---



_Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/sillsdev/chorus/81)_

PR #83, opened 2016-02-18
Fix ServerSettingsModel.InitFromUri to use UrlDecode

megahirt:
This fixes a bug in the Chorus Internet Settings Dialog which was reported to me by a WeSay user.  The previous code would allow a situation whereby the user has a password with a special character in it like % that when saved into the password field and then the dialog is re-opened, the password has "grown" in length.  This is caused by the password field not being properly url decoded when extracted from the URL.

A regression test is also accompanied.

You can test this bug in the latest version of either Fieldworks, WeSay or any other app using the Chorus Internet Settings Dialog box.

[Review on Reviewable](https://reviewable.io/reviews/sillsdev/chorus/83)

### DIFF
--- a/src/LibChorus/Model/ServerSettingsModel.cs
+++ b/src/LibChorus/Model/ServerSettingsModel.cs
@@ -54,9 +54,9 @@ namespace Chorus.Model
 		public virtual void InitFromUri(string url)
 		{
 			SetServerLabelFromUrl(url);
-			Password = UrlHelper.GetPassword(url);
-			AccountName = UrlHelper.GetUserName(url);
-			ProjectId = UrlHelper.GetPathAfterHost(url);
+			Password = HttpUtilityFromMono.UrlDecode(UrlHelper.GetPassword(url));
+			AccountName = HttpUtilityFromMono.UrlDecode(UrlHelper.GetUserName(url));
+			ProjectId = HttpUtilityFromMono.UrlDecode(UrlHelper.GetPathAfterHost(url));
 			CustomUrl = UrlHelper.GetPathOnly(url);
 			//CustomUrlSelected = true;
 		}
--- a/src/LibChorusTests/Model/ServerSettingsModelTests.cs
+++ b/src/LibChorusTests/Model/ServerSettingsModelTests.cs
@@ -1,4 +1,5 @@
 ﻿using System.IO;
+using System.Security.Principal;
 using NUnit.Framework;
 using SIL.Progress;
 using SIL.TestUtilities;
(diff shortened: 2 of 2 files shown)


ermshiperete:
Two suggestions in the test, otherwise LGTM.

---

Reviewed 2 of 2 files at r1.
Review status: all files reviewed at latest revision, 2 unresolved discussions, some commit checks failed.

---

*[src/LibChorusTests/Model/ServerSettingsModelTests.cs, line 25 \[r1\]](https://reviewable.io:443/reviews/sillsdev/chorus/83#-KAnGo0l04OxZifkBxv0:-KAnGo0l04OxZifkBxv1:1350486327) ([raw file](https://github.com/sillsdev/chorus/blob/91fa455216e9265fe47519750d395314ea4d38ac/src/LibChorusTests/Model/ServerSettingsModelTests.cs#L25)):*
Using `model` instead of `m` would make it a little clearer, but I see that's the name that is used on other tests as well.

---

*[src/LibChorusTests/Model/ServerSettingsModelTests.cs, line 33 \[r1\]](https://reviewable.io:443/reviews/sillsdev/chorus/83#-KAnH8MWT0OnKltn9mzN:-KAnH8MX1fLubOkT47fp:12996244) ([raw file](https://github.com/sillsdev/chorus/blob/91fa455216e9265fe47519750d395314ea4d38ac/src/LibChorusTests/Model/ServerSettingsModelTests.cs#L33)):*
I didn't look at the internals of `ServerSettingsModel`, but it looks like there are two ways to initialize properties: by setting the individual properties (which then creates the `URL`), or by calling `InitFromUri` with a URL. 

If that's correct then it would be better to assign a new `ServerSettingsModel` instance to `m` before calling `InitFromUri`.

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/sillsdev/chorus/83#-:-KAnI-CTSBuVRNjj4N2w:-1090485592)*

megahirt:
This bug could potentially affect any user having a password that contains special characters like **%** or **@** or **&**  After initially entering a password into the dialog having a special character like **&**, the user is able to do send/receive.  For example:

image

Typically, once account information is entered by the user, the settings dialog closed, and the user successfully does a send/receive, the user never has reason to go into the settings dialog again.  If however, they do enter the settings dialog again, the password will have seemingly grown by several characters, for example as seen below:

image

One only needs to save and re-open the settings dialog and the password continues to grow (send/receive is not necessary - in fact it will fail with the incorrect password):

image

and grow:

image

ermshiperete:
Hey, that's not a bug, that's a feature - it increases password security :wink:

ermshiperete:
:LGTM:

---

Reviewed 1 of 1 files at r2.
Review status: all files reviewed at latest revision, 2 unresolved discussions, some commit checks failed.

---



*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/sillsdev/chorus/83#-:-KAnKrZVqBCQv0WB5j46:1386550993)*

ermshiperete:
:lgtm: - oh, lgtm needs to be typed as lowercase to show up as icon 

---

Review status: all files reviewed at latest revision, 2 unresolved discussions, some commit checks failed.

---



*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/sillsdev/chorus/83#-:-KAnL-E69hRgFeNMJKoM:-99808207)*

megahirt:
Review status: all files reviewed at latest revision, 2 unresolved discussions, some commit checks failed.

---

*[src/LibChorusTests/Model/ServerSettingsModelTests.cs, line 33 \[r1\]](https://reviewable.io:443/reviews/sillsdev/chorus/83#-KAnH8MWT0OnKltn9mzN:-KAnN3pm4LR__fOxdayO:2019904527) ([raw file](https://github.com/sillsdev/chorus/blob/91fa455216e9265fe47519750d395314ea4d38ac/src/LibChorusTests/Model/ServerSettingsModelTests.cs#L33)):*
thanks for the good suggestions.  I will modify my test to suit.

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/sillsdev/chorus/83)*

megahirt:
There was a bit of palaso bob build confusion as my pull request unnecessarily triggered a build on the chorus-2.5 branch and the lfmerge branch, which are unrelated to this PR.  I learned that these two build targets were also polling the refs/heads/pulls github branch and they should not, so I removed those from the TC build configuration.  Now, only TC configs tracking master will trigger PR builds, and I believe that's the way it should be.

ermshiperete:
Reviewed 1 of 1 files at r3.
Review status: all files reviewed at latest revision, 1 unresolved discussion.

---



*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/sillsdev/chorus/83)*

jasonleenaylor:
Due to my new PR I noticed this one languishing here. It looks like it might be ready to merge. The bugfix looks good to me. Is there anything more to do here? @ermshiperete @megahirt

PR #85, opened 2016-03-24
Make ChorusSystem more UI independent (v2)

ermshiperete:
This was done by splitting out a UI independent `ChorusSystemSimple` base class and is needed for LfMerge.

Also use Autofac from nuget package.

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/chorus/85)


sillsdevgerrit:
Build finished.

sillsdevgerrit:
Build finished.

sillsdevgerrit:
Build finished.

sillsdevgerrit:
Build finished.

sillsdevgerrit:
Build finished.

sillsdevgerrit:
Build finished.

sillsdevgerrit:
Build finished.

sillsdevgerrit:
Build finished.

sillsdevgerrit:
Build finished.

sillsdevgerrit:
Build finished.

regnrand:
I did some high level reviewing. The biggest issue is exposing Autofac to Chorus clients in IChorusSystem. I'm less sure there is a need for a UI-less version of ChorusSystem and a lot of changes seem to support that. The new class "SyncControlModelSimple" has some merit, but it isn't usable by clients yet that need to use an adjunct other than the default one in Chorus. (No way yet to get one added to "SyncControlModelSimple".)

---

Review status: 0 of 37 files reviewed at latest revision, 2 unresolved discussions, some commit checks failed.

---

*[src/LibChorus/IChorusSystem.cs, line 23 \[r6\]](https://reviewable.io:443/reviews/sillsdev/chorus/85#-KFQvOGmdgaLGhH9tuCS:-KFQvOGnylDv83iV0BuQ:2133613751) ([raw file](https://github.com/sillsdev/chorus/blob/1e4cce13f0b845f367e8c8e394a2eabae7f36723/src/LibChorus/IChorusSystem.cs#L23)):*
Chorus would really prefer to get rid of its use of Autofac altogether in favor of MEF (isn't possible with MEF 1, however). We don't really don't want to spread the dependency on Autofac, which this property would require.

---

*[src/LibChorus/sync/SyncControlModelSimple.cs, line 19 \[r6\]](https://reviewable.io:443/reviews/sillsdev/chorus/85#-KFQzvzrrm2GFpfrrEgt:-KFQzvzrrm2GFpfrrEgu:-1409630053) ([raw file](https://github.com/sillsdev/chorus/blob/1e4cce13f0b845f367e8c8e394a2eabae7f36723/src/LibChorus/sync/SyncControlModelSimple.cs#L19)):*
I thought I saw a need for this class in Flex Bridge, but I see it creates one but then only uses its "SyncOptions" property. Is there other justification to make this new class?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/sillsdev/chorus/85#-:-KFW7CGiE0KXh7CAI72T:-1014475942)*

PR #110, opened 2017-01-06
Inform users about corrupt USB Repos instead of crashing

jasonleenaylor:
With this change corrupt repositories on a USB drive appear as follows:

image

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/chorus/110)


jasonleenaylor:
The tooltip message has been updated to: "The folder {0} can not be used for Send/Receive. It may be corrupt." for increased accuracy about what we are detecting.

PR #112, opened 2017-02-03
Allow specifying annotation GUID in constructor

rmunn:
This is needed for Language Forge Send/Receive, where comments created in Language Forge will have pre-existing GUIDs and we will want to create Annotations with the same GUID. This way, the annotation in Chorus Notes and the comment in Language Forge will be trackable as being the same object, and if the comment is resolved in FLEx then the corresponding comment in Language Forge can be marked as closed.

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/chorus/112)

### DIFF
--- a/src/LibChorus/notes/Annotation.cs
+++ b/src/LibChorus/notes/Annotation.cs
@@ -29,10 +29,15 @@ namespace Chorus.notes
         }
 
 		public Annotation(string annotationClass, string refUrl, string path)
+			: this(annotationClass, refUrl, System.Guid.NewGuid(), path)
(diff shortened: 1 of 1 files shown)


ermshiperete:
:lgtm:

The TC builds failed because they tried to build on the wrong branch.

---

Reviewed 1 of 1 files at r1.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/sillsdev/chorus/112#-:-Kc20r97h6R48C0GV2cV:b-eg4f2z)*

PR #113, opened 2017-02-17
Avoid too-long filename (LT-18093)

papeh:
Omit blatantly redundant parts of a150-character filename.
TODO in a follow-up: remove further name parts to get a fixed-length
name

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/chorus/113)

### DIFF
--- a/src/LibChorus/VcsDrivers/Mercurial/HgResumeTransport.cs
+++ b/src/LibChorus/VcsDrivers/Mercurial/HgResumeTransport.cs
@@ -221,7 +221,6 @@ namespace Chorus.VcsDrivers.Mercurial
 			return commonBases;
 		}
 
-
 		private MultiMap<string, string> GetRemoteRevisions(int offset, int quantity)
 		{
 			const int totalNumOfAttempts = 2;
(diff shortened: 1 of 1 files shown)


papeh:
@palaso-bob, this PR targets `chorus-2.5`, but was built using the `master` configuration, so the checks are invalid.

src/LibChorus/VcsDrivers/Mercurial/HgResumeTransport.cs
```diff
@@ -320,12 +320,8 @@ public void Push()
...
 
 			// create a bundle to push
-			string tip = _repo.GetTip().Number.Hash;
-			string bundleId = "";
-			foreach (var revision in baseRevisions)
-			{
-				bundleId += String.Format("{0}-{1}", revision.Number.Hash, tip);
-			}
+			var tip = _repo.GetTip().Number.Hash;
+			var bundleId = string.Join("_", baseRevisions.Select(rev => rev.Number.Hash)) + '-' +  tip; // REVIEW (Hasso) 2017.02: how much of this is necessary? It's too long when used as a filename. (see also line 639)
```
jasonleenaylor: Does the revision number have a short version of the hash? If so I think we could use that safely. If not LGTM.

jasonleenaylor:
One comment to consider. LGTM

src/LibChorus/VcsDrivers/Mercurial/HgResumeTransport.cs
papeh: we could call Substring

PR #115, opened 2017-02-17
Clean up Bundle ID

papeh:
* Omit all Base Revision hashes from Bundle ID.
  The last Base Revision hash is already in the directory path (in most cases), and the files are deleted immetiately on a successful S/R, so the Tip hash should be sufficiently unambiguous.
* Clean up the internal BundleStorageManager API.

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/chorus/115), and should be reviewed by someone who's worked on something other than FLEx.

### DIFF
--- a/src/LibChorus/VcsDrivers/Mercurial/BundleStorageManager.cs
+++ b/src/LibChorus/VcsDrivers/Mercurial/BundleStorageManager.cs
@@ -3,34 +3,24 @@ using System.IO;
 
 namespace Chorus.VcsDrivers.Mercurial
 {
-	abstract class BundleStorageManager
+	internal abstract class BundleStorageManager
 	{
 		public string BundlePath;
-		private const int _numberOfHoursToKeepIncompleteData = 24;
-		private string _dataFolderPath;
-		private string _storagePath;
-		private string _bundleId;
+		private const int NumberOfHoursToKeepIncompleteData = 24;
 		private string _idFilePath;
 		public readonly string TransactionId;
 
-		public virtual string StorageFolderName
+		protected BundleStorageManager(string storagePath, string storageFolderName, string bundleIdFilename)
 		{
-			get { return "overrideme"; }
-		}
+			var dataFolderPath = Path.Combine(storagePath, storageFolderName);
 
-		public BundleStorageManager(string storagePath, string bundleId)
-		{
-			_storagePath = storagePath;
-			_bundleId = bundleId;
-			_dataFolderPath = Path.Combine(storagePath, StorageFolderName);
+			Directory.CreateDirectory(storagePath); // create if necessary
+			Directory.CreateDirectory(dataFolderPath); // create if necessary
 
-			Directory.CreateDirectory(_storagePath); // create if necessary
-			Directory.CreateDirectory(_dataFolderPath); // create if necessary
+			CleanUpExpiredData(dataFolderPath);
 
-			CleanUpExpiredData();
-
-			TransactionId = GetTransactionId();
-			BundlePath = Path.Combine(_dataFolderPath, String.Format("{0}.bundle", TransactionId));
+			TransactionId = GetTransactionId(dataFolderPath, bundleIdFilename);
+			BundlePath = Path.Combine(dataFolderPath, String.Format("{0}.bundle", TransactionId));
 			if (!File.Exists(BundlePath))
 			{
 				var fs = new FileInfo(BundlePath).Create();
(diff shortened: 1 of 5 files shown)


megahirt:
I'm not convinced that a bundleId based solely on tip is actually unique enough.  See comments below.

---







*Comments from [Reviewable](https://reviewable.io:443/reviews/sillsdev/chorus/115#-:-KdcuyieA6RUrGQlNDWa:bod6d44)*

megahirt:
*[src/LibChorus/VcsDrivers/Mercurial/HgResumeTransport.cs, line 322 at r1](https://reviewable.io:443/reviews/sillsdev/chorus/115#-Kdcqe7mwbDmPSqFWxO_:-Kdcqe7mwbDmPSqFWxOa:b-i4q2vt) ([raw file](https://github.com/sillsdev/chorus/blob/276f5387bdd3fd30eff66620c5864a4bdf7f9504/src/LibChorus/VcsDrivers/Mercurial/HgResumeTransport.cs#L322)):*
> ```Smalltalk
> 
> 			// create a bundle to push
> 			var bundleHelper = new PushStorageManager(PathToLocalStorage, "push" + _repo.GetTip().Number.Hash);
> ```

I'm not convinced that we can uniquely identify a bundle by the tip of the local repo.  A bundle is fundamentally a collection of changesets between two revisions.  In the push case, we have more revisions than the server, and so the bundle consists of a range of revisions up to and including tip.  If the bundleId is based solely on the local tip, then I could imagine a scenario where two different bundles have the same bundleId (because the the local repo is still at the same tip but the server's common base hashes have changed).  My recommendation for a shortened bundleId is something like:

`var bundleId = _repo.GetTip().Number.Hash + '_' + baseRevisions.First()`

(I can't tell at the moment if it should be baseRevisions.First() or baseRevisions.Last() to get the oldest revision)

This would ensure that the bundleFileNameId accurately represents the uniqueness of the underlying data, spanning from the first common revision to the local tip.  Just using tip() in the name doesn't guarantee that.

One thing is certain though, including the hash of every commit in the bundle is overkill and unnecessary.

If the bundleId got confused with a different bundle doing a Push operation, the worst that could happen is that data is unnecessarily sent the server (and discarded as redundant).  During a client Pull operation, I'm not certain what the consequences would be for a non-unique bundleId filename.  I'd have to think about that longer.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/sillsdev/chorus/115)*

ermshiperete:
Review status: 0 of 5 files reviewed at latest revision, 1 unresolved discussion, some commit checks failed.

---

*[src/LibChorus/VcsDrivers/Mercurial/HgResumeTransport.cs, line 322 at r1](https://reviewable.io:443/reviews/sillsdev/chorus/115#-Kdcqe7mwbDmPSqFWxO_:-KddnhRedxfmNE89yt1I:b-nfseyj) ([raw file](https://github.com/sillsdev/chorus/blob/276f5387bdd3fd30eff66620c5864a4bdf7f9504/src/LibChorus/VcsDrivers/Mercurial/HgResumeTransport.cs#L322)):*
<details><summary><i>Previously, megahirt (Chris Hirt) wrote…</i></summary><blockquote>

I'm not convinced that we can uniquely identify a bundle by the tip of the local repo.  A bundle is fundamentally a collection of changesets between two revisions.  In the push case, we have more revisions than the server, and so the bundle consists of a range of revisions up to and including tip.  If the bundleId is based solely on the local tip, then I could imagine a scenario where two different bundles have the same bundleId (because the the local repo is still at the same tip but the server's common base hashes have changed).  My recommendation for a shortened bundleId is something like:

`var bundleId = _repo.GetTip().Number.Hash + '_' + baseRevisions.First()`

(I can't tell at the moment if it should be baseRevisions.First() or baseRevisions.Last() to get the oldest revision)

This would ensure that the bundleFileNameId accurately represents the uniqueness of the underlying data, spanning from the first common revision to the local tip.  Just using tip() in the name doesn't guarantee that.

One thing is certain though, including the hash of every commit in the bundle is overkill and unnecessary.

If the bundleId got confused with a different bundle doing a Push operation, the worst that could happen is that data is unnecessarily sent the server (and discarded as redundant).  During a client Pull operation, I'm not certain what the consequences would be for a non-unique bundleId filename.  I'd have to think about that longer.
</blockquote></details>

Another option would be to create the name from the hash value over (baseTip, localTip) (like Chris' suggestion above), or could also be done over all included commits.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/sillsdev/chorus/115)*

papeh:
@jasonleenaylor's comments in FlowDock: The real answer to if it is safe comes from what happens to the file on the server. Hasso's changes are certainly safe on the client. Is the filename even retained in the trip through the resumable API?

My response: The Bundle ID is used locally to identify a file that contains the Transaction ID (which is a GUID that is generated each time the file is created. The Transaction ID is passed as a parameter to the HG resumeable API. The file containing the Transaction ID is deleted promptly on the completion of the transaction. (I could have been clearer about that in my commit message)

PR #118, opened 2017-03-30
Don't modify SourcesToTry when iterating over them

papeh:
LT-18276, which we cannot reproduce

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/chorus/118)


jasonleenaylor:
There is a race condition due to a separate thread that updates network sources. This prevents the crash. In the case where it would have crashed I expect it will do nothing, which is fine.

PR #120, opened 2017-04-20
Upgrade to .Net 4.6.1

jwickberg:
This pull request contains the changes needed to upgrade to .Net 4.6.1

Not positive that this should be put on the master branch - seems like the best place but it means other apps using the master builds will need to upgrade also.

I made a change to the WIX project so that the ChorusHub installer will depend on 4.6.1 or later being installed - this means that WIX 3.11 RC needs to be used for building the installer, at least I didn't see any support for .Net 4.6.1 in the WIX 3.10 tool set.

Until the build agents are updated, the ChorusHub installer will fail to build.

I tried upgrading the NUnit version, but ran into problems because Palaso library using NUnit 2.6.4. So I'm not sure if the tests are actually being run using .Net 4.6.1

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/chorus/120)


jwickberg:
Will rework a couple of items for Linux and look into the test failures

PR #123, opened 2017-05-02
Upgrade to .Net 4.6.1

jwickberg:
This pull request contains the changes needed to upgrade to .Net 4.6.1

Not positive that this should be put on the master branch - seems like the best place but it means other apps using the master builds will need to upgrade also.

I made a change to the WIX project so that the ChorusHub installer will depend on 4.6.1 or later being installed - this means that WIX 3.11 RC needs to be used for building the installer, at least I didn't see any support for .Net 4.6.1 in the WIX 3.10 tool set.

Also change LanguageVersion setting to be C#6 so we won't use C#7 as the default when VS2017 is used.

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/chorus/123)


jwickberg:
I tried running the failed test with Resharper in VS and it passes, so I'm not sure what to do about that test.

jasonleenaylor:
The work we did in L10NSharp and that is still ongoing for libpalaso to get tests running in mono4-sil will be needed for the chorus build as well I think. I'm guessing that the windows test failure might be intermittent, so if you can fix it awesome!

FoolRunning:
The only thing that I see that might need fixing is that there are a bunch of tabs that were converted to spaces in some of the files.

andrew-polk:
What is the migration strategy for adopting this? Do we need to bump the version number? Do we need to create a duplicate set of TC builds? How do we ensure clients don't break?
See https://www.flowdock.com/app/lsdev/commons/threads/3YZONhumY91-sxHiKFp3PkwBwOy for what I did in relation to these questions for libpalaso.

jasonleenaylor:
Superceeded by PR 132

PR #125, opened 2017-06-21
Always write timestamps with time zones

rmunn:
In some cases, `TimeFormatNoTimeZone` was lying, writing out a local time with a "Z" suffix, which means "This is UTC". This results in loss of data, since the *actual* time of the original event (conflict, message, etc.) can no longer be retrieved. If it's important to write out local time rather than UTC in a timestamp, then that timestamp should carry the actual time zone it was created in, rather than a false Z suffix.

To ensure that other code calling the Chorus library won't break, I haven't removed the `TimeFormatNoTimeZone` constant (though I was tempted to). Instead, I have marked it Obsolete and suggested the correct constant (`TimeFormatWithTimeZone`) be used instead.

This change passes all Chorus unit tests that I'm aware of, but I was not able to test it with every piece of software that might use Chorus.

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/chorus/125)

### DIFF
--- a/src/LibChorus/merge/xml/generic/Conflict.cs
+++ b/src/LibChorus/merge/xml/generic/Conflict.cs
@@ -44,7 +44,9 @@ namespace Chorus.merge.xml.generic
 	/// </summary>
 	public abstract class Conflict : IConflict, IEquatable<Conflict>
 	{
+		[Obsolete("Use TimeFormatWithTimeZone instead, as TimeFormatNoTimeZone produces incorrect results when used with DateTime.Now")]
 		static public string TimeFormatNoTimeZone = @"yyyy-MM-ddTHH:mm:ssZ";
+		static public string TimeFormatWithTimeZone = @"yyyy-MM-ddTHH:mm:ssK";
 
 		private ContextDescriptor _context = new NullContextDescriptor();
 	   // protected string _shortDataDescription;
@@ -152,7 +154,7 @@ namespace Chorus.merge.xml.generic
 			writer.WriteAttributeString(@"author", string.Empty, Author);
 			writer.WriteAttributeString(@"status", string.Empty, @"open");
 			writer.WriteAttributeString(@"guid", string.Empty, Guid.ToString());//nb: ok to have the same guid with the conflict, as they are in 1-1 relation and eventually we'll remove the one on conflict
-			writer.WriteAttributeString(@"date", string.Empty, DateTime.UtcNow.ToString(TimeFormatNoTimeZone));
+			writer.WriteAttributeString(@"date", string.Empty, DateTime.UtcNow.ToString(TimeFormatWithTimeZone));
 			writer.WriteString(GetFullHumanReadableDescription());
 
 			//we embed this xml inside the CDATA section so that it pass a more generic schema without
@@ -192,7 +194,7 @@ namespace Chorus.merge.xml.generic
 			//writer.WriteAttributeString("pathToUnitOfConflict", string.Empty, PathToUnitOfConflict);
 			writer.WriteAttributeString(@"type", string.Empty, Description);
 			writer.WriteAttributeString(@"guid", string.Empty, Guid.ToString());
-			writer.WriteAttributeString(@"date", string.Empty, DateTime.UtcNow.ToString(TimeFormatNoTimeZone));
+			writer.WriteAttributeString(@"date", string.Empty, DateTime.UtcNow.ToString(TimeFormatWithTimeZone));
 		  //  writer.WriteAttributeString("shortDataDescription", _shortDataDescription);
 			writer.WriteAttributeString(@"whoWon", _whoWon);
 			writer.WriteAttributeString(@"htmlDetails", HtmlDetails);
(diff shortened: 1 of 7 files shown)


rmunn:
Looks like the TeamCity build has some problems unrelated to this PR: it's complaining about not finding the Palaso library. Something to look into, but not something that should block anyone from reviewing this PR.

ermshiperete:
One question I have is if it is really necessary to define identical `TimeFormatWithTimeZone` strings so many times? I know that was already there, but since we're touching it we could probably do some cleanup.

This change will modify the way most timestamps look: instead of `2017-06-21T10:36:00Z` we now get `2017-06-21T10:36:00+02:00`. Will this cause problems with clients? Or do they parse the dates in a way that understands the timezones?

If we'd want to maintain compatibility of the file format we could change the calls of `DateTime.Now` to `DateTime.UtcNow`. Although I just saw your message on flowdock that the use of local time was intentional.

And finally the question (not directly related to this change again): do the time format strings have to be `public` or could they be made `internal`?

---

Reviewed 7 of 7 files at r1.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/sillsdev/chorus/125#-:-Kn8mGZV8_mVWLUMwjrF:bi6wjd7)*

rmunn:
If we're going to change where `TimeFormatWithTimeZone` is defined, it probably makes more sense to move it into a lower-level library like LibPalaso than to keep it in Chorus. But that's an even bigger change and would require touching two libraries that almost everyone uses instead of just one. So I'd prefer to do that as a separate step.

That change is why I want to proceed cautiously. I'd guess that most software that's parsing timestamps will not be parsing it with a regex, but with something like `DateTime.TryParse`. If so, then they should handle timestamps with time zones just fine. But if there's any software that's parsing a timestamp from a Chorus file with a regex, and *counting* on the Z at the end in its regex, then that could break. I'd consider that a bug in that software and change it to use DateTime.TryParse, but I can't guarantee that that won't happen. When we do merge this, we'll need to hunt through our existing software to find where Chorus-generated files are being parsed, and what is done with timestamps in those cases.

Yes, changing it to UtcNow would be the simpler solution, and is the first one I implemented until I learned the reason for that decision.

Some of these time format strings may be accessed by client software, which is why I decided not to delete the incorrent `TimeFormatNoTimeZone` string and mark it Obsolete instead. That way we'll only produce compiler warnings, not broken builds, when we merge this in. If we merge this in and find out that absolutely *nothing* outside of Chorus itself is using these strings, then we can mark then `internal`.

---

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/sillsdev/chorus/125#-:-Kn90Ba0_sJD8wR4X0CQ:b4hzk1t)*

src/LibChorus/merge/xml/generic/Conflict.cs
```diff
@@ -152,7 +154,7 @@ public void WriteAsChorusNotesAnnotation(XmlWriter writer)
 			writer.WriteAttributeString(@"author", string.Empty, Author);
 			writer.WriteAttributeString(@"status", string.Empty, @"open");
 			writer.WriteAttributeString(@"guid", string.Empty, Guid.ToString());//nb: ok to have the same guid with the conflict, as they are in 1-1 relation and eventually we'll remove the one on conflict
-			writer.WriteAttributeString(@"date", string.Empty, DateTime.UtcNow.ToString(TimeFormatNoTimeZone));
+			writer.WriteAttributeString(@"date", string.Empty, DateTime.UtcNow.ToString(TimeFormatWithTimeZone));
```
papeh: Is the intent to go to all local time? This file still uses DateTime.UtcNow throughout.

rmunn:
Rebasing this old PR since it's still needed and hasn't been looked at in 7 months.

rmunn:
To answer @papeh's question, I don't know why whoever wrote that line chose to use UtcNow rather than Now, and I'd rather not change it until/unless I know why it was written that way in the first place. My intent with this PR is to fix a long-standing bug where Chorus produces false timestamps. There's at least one feature (the Language Forge Send/Receive feature where we are now trying to merge comments as well) that is being hindered by the inability to rely on the timestamps coming from Chorus. (If we trust the timestamps that Chorus produces, then we see things like "Robin Munn added a comment 7 hours from now" in the Language Forge comments: I made that comment in FLEx at 10:00 AM local time, in the UTC+7 time zone. But it was recorded as being 10:00 AM UTC, which is 17:00 PM local time — and so the comment appears to be "from the future". Since one of the things we need to do in the Send/Receive feature is determine which comment is the *most recent* comment in this discussion thread, that causes problems.)

ermshiperete:
:lgtm:

---

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/sillsdev/chorus/125#-:-L4eEesD_OAB9kax6d7F:bnfp4nl)*

jasonleenaylor:
So long as the reading of these timezones won't be a burden to each client i'm satisfied.

jasonleenaylor:
I'm leaving this merge up to someone who will be responsible for the result. @ermshiperete or @rmunn can merge this one in.

rmunn:
Resolved conflicts so this is mergeable again. I'll merge it once the Jenkins build complete.

PR #131, opened 2017-07-18
Fix : LT-18129 Crash on Conflict Details dialog

samdoss:
Fix : LT-18129 Crash on Conflict Details dialog
  * Removed the WebBrowser Copy execution
  * Included the Clipboard copy function
  * Now it will work on linux also.

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/chorus/131)


jasonleenaylor:
The builds did not run against chorus-2.5. Looks like the PR system needs some tweaking.
Locally verified the windows build before merging.

PR #132, opened 2017-07-18
Upgrade to .Net 4.6.1

andrew-polk:
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/chorus/132)


JohnThomson:
:lgtm:. Rather wish we could remove the changes where files just lost their final newline, but probably more trouble than it's worth. One minor question on a version number. It's not clear to me how the added <Prefer32Bit>false</Prefer32Bit> elements relate...I'm guessing it's something VS did while you weren't looking...probably harmless

---

Reviewed 23 of 26 files at r1, 3 of 3 files at r2.

---

*[src/Installer/ChorusHub.wxs, line 4 at r2](https://reviewable.io:443/reviews/sillsdev/chorus/132#-KpKytIbD9ifcZGLTuw1:-KpKytIbD9ifcZGLTuw2:bzg7e19) ([raw file](https://github.com/sillsdev/chorus/blob/631f06c47dfada9dac4ed6a35379eb5720d24816/src/Installer/ChorusHub.wxs#L4)):*
> ```XML
> 
> 
> <?define Property_ProductVersion = "2.6.3.0" ?>
> ```

Was this meant to be left for a build script to replace?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/sillsdev/chorus/132#-:-KpKzOgvlqIOrTIjmyO0:bbm568q)*

PR #137, opened 2017-09-25
fix mercurial repo tests on windows

glasseyes:
the tests create an empty file for the mercurial binary in the test dir but hard coded hg instead of using hg.exe on windows

missing eol extension in the test ini file data on windows

The change to RepositoryRecoversFromIncompleteMerge may not be necessary but allows it to work with older builds of libpalaso than the one that will come from the libpalaso PR

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/chorus/137)

### DIFF
--- a/src/LibChorusTests/VcsDrivers/Mercurial/RepositoryTests.cs
+++ b/src/LibChorusTests/VcsDrivers/Mercurial/RepositoryTests.cs
@@ -2,12 +2,12 @@ using System;
 using System.IO;
 using System.Reflection;
 using Chorus;
-using Chorus.VcsDrivers;
 using Chorus.VcsDrivers.Mercurial;
 using Ionic.Zip;
 using LibChorus.TestUtilities;
 using NUnit.Framework;
 using Palaso.IO;
+using Palaso.PlatformUtilities;
 using Palaso.Progress;
 using Palaso.TestUtilities;
 
@@ -129,7 +129,7 @@ namespace LibChorus.Tests.VcsDrivers.Mercurial
 		{
 			using (var tempRepo = new TemporaryFolder("ChorusIncompleteMerge"))
 			{
-				var baseDir = Path.GetDirectoryName(Assembly.GetExecutingAssembly().CodeBase);
+				var baseDir = FileUtils.NormalizePath(Path.GetDirectoryName(Assembly.GetExecutingAssembly().CodeBase));
 				baseDir = FileUtils.StripFilePrefix(baseDir);
 				var zipFile = new ZipFile(Path.Combine(baseDir, Path.Combine("VcsDrivers", Path.Combine("TestData", "incompletemergerepo.zip"))));
 				zipFile.ExtractAll(tempRepo.Path);
@@ -165,10 +165,18 @@ namespace LibChorus.Tests.VcsDrivers.Mercurial
 
 		private static string GetExtensionsSection(string pathToMercurialFolder)
 		{
+#if !MONO
+			return string.Format(@"[extensions]
+eol=
+hgext.graphlog=
+convert=
+fixutf8={0}\MercurialExtensions\fixutf8\fixutf8.py", Path.GetDirectoryName(pathToMercurialFolder));
+#else
 			return string.Format(@"[extensions]
 hgext.graphlog=
 convert=
 fixutf8={0}/MercurialExtensions/fixutf8/fixutf8.py", Path.GetDirectoryName(pathToMercurialFolder));
+#endif
 		}
 
 		[Test]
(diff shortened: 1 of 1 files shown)


src/LibChorusTests/VcsDrivers/Mercurial/RepositoryTests.cs
```diff
@@ -165,10 +165,18 @@ public void CloneToUsbWithoutUpdateFollowedByIdentifierDoesNotAffectHgrc()
 
 		private static string GetExtensionsSection(string pathToMercurialFolder)
 		{
+#if !MONO
```
jasonleenaylor: I'm alarmed here. So the two platforms should have the same extensions. The data can change (for path purposes) but I don't think we can have users on different OS work together if we can't have the same extensions.

jasonleenaylor:
There was a point where all these tests were passing without these changes right? I'm not sure why the extensions need to be modified, and I'm not sure that that is correct. I can't remember if Chorus 2.5 is running with a mercurial that needs that extension.

glasseyes:
Alas not, the continuous chorus 2.5 win32 build has been failing since these tests were committed on Sep 4. Firstly the tests needed the fw-8.3.5 build of libpalaso rather than fw-8.3.1, then the issues fixed in this PR. 

---

Review status: 0 of 1 files reviewed at latest revision, all discussions resolved, some commit checks failed.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/sillsdev/chorus/137#-:-KutHe4EYQk-yhYHHtcb:b5usyx6)*

glasseyes:
I changed what the test expects in terms of extensions to match what is in src/LibChorus/VcsDrivers/Mercurial/HgRepository.cs git blame says we were both a part of that ;) but I can't remember the details now

---

Review status: 0 of 1 files reviewed at latest revision, all discussions resolved, some commit checks failed.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/sillsdev/chorus/137#-:-KutMSC00nO_5BfZDl5w:bes6530)*

rmunn:
I'll address Jason's comments about the extensions first, then I'll review Daniel's code.

I remember a problem that Eberhard and I helped Daniel look into around the start of September, where it turned out that the problem was an extra `eol` line that the package installer (for WeSay, IIRC) had added to the `.hgrc`. Which was fine on Windows, but was failing on Linux.

I'm not particularly happy with the idea that we have a cross-platform library that *requires* separate builds for Windows and Linux because it expects different data on each platform. What I said on Flowdock at the time was:

> ... an alternate API might be to add a third argument [to the `CheckExtensions` method], which is a dictionary of "optional extensions". They're allowed to be absent, but if present they should have these values. That way the eol extension can be checked for without messing up the LibChorus-consuming programs that don't want the eol extension.

I still think that's a good idea. If the `eol` extension is present in the `.hgrc` file that is found on a Linux system, I don't think we should throw an exception since it will do no harm on Linux (it will convert `\n` characters to `\n`). And what's especially bad is line 230 of `HgRepository.cs`, which can end up lying to you about the cause of the problem (it ALWAYS says that the problem is that `fixutf8` was absent, when in fact the problem can sometimes be that `eol` was *present* and not expected). That caused Daniel, Eberhad and me to spend several hours hunting down the problem a couple of weeks ago, because the error message sent us hunting in the wrong direction at first.

Okay, now I'll go ahead and review Daniel's code.

rmunn:
:lgtm:, though as I said before I'd really like the `eol` extension to NOT cause a failure if it's present on a Linux system. So after this PR is merged, a second one would be good to fix that issue.

---

Review status: 0 of 1 files reviewed at latest revision, 1 unresolved discussion, some commit checks failed.

---

*[src/LibChorusTests/VcsDrivers/Mercurial/RepositoryTests.cs, line 168 at r1](https://reviewable.io:443/reviews/sillsdev/chorus/137#-KuxFn1iDNLFNBzGbE7Q:-KuxFn1iDNLFNBzGbE7R:b66awjw) ([raw file](https://github.com/sillsdev/chorus/blob/152c5149f2c5549291ef8bedfa9c109ff4edd8d8/src/LibChorusTests/VcsDrivers/Mercurial/RepositoryTests.cs#L168)):*
> ```Smalltalk
> 		private static string GetExtensionsSection(string pathToMercurialFolder)
> 		{
> #if !MONO
> ```

I see from a quick search through the rest of LibChorus that we tend to use `#if MONO ... #else ... #endif` rather than `#if !MONO ... #else ... #endif`. I'd prefer reordering this to match, so that the "true" case of the `#if` comes first rather than the "false" case.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/sillsdev/chorus/137#-:-KuxGhTYeIyeZ-K8xAB4:bsql7u9)*

rmunn:
https://github.com/sillsdev/chorus/commit/05d838763f39125ec62a11576c488827ef120add is when the `eol` extension was removed on Mono (though it was called the `hgext.win32text` extension at the time). I haven't yet traced that commit's history to see which branch(es) it ended up in besides `master`.

rmunn:
Reviewed 1 of 1 files at r1.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/sillsdev/chorus/137)*

PR #144, opened 2017-11-02
Correct project TC bt to master

marksvc:
Set TC build_type to the TC job building on master, to match this
branch. Then we will download the correct dependencies and build
according to APIs that clients are expecting us to be using.

For example, recognize that SIL.UsbDrive types were moved from
SIL.Core.dll to SIL.Core.Deskotp.dll.

This mimics what was done in buildupdate.win.sh in recent commit
13c964894.

----
I'm not used to doing this, so I will be glad for someone else to say this was the right thing to do. Also, this might be introducing some unit test failures, but I'm not sure if my workstation was set up correctly to test.

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/chorus/144)


glasseyes:
:lgtm:

---

Review status: 0 of 1 files reviewed at latest revision, all discussions resolved, some commit checks failed.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/sillsdev/chorus/144#-:-KxzcMG_Yik7zJtmD2qo:bnfp4nl)*

PR #154, opened 2017-12-08
Fix LT-14385 Store all valid merge candidate childnodes.

RaymondLuong3:
* All childnodes of a parent will be checked for matching key

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/chorus/154)

### DIFF
--- a/src/LibChorus/merge/xml/generic/FindNodeToMerge.cs
+++ b/src/LibChorus/merge/xml/generic/FindNodeToMerge.cs
@@ -76,7 +76,11 @@ namespace Chorus.merge.xml.generic
 		private readonly string _keyAttribute;
 #if USEOPTIMIZEDVERSION
 		private readonly List<XmlNode> _parentsToSearchIn = new List<XmlNode>();
-		private readonly Dictionary<int, Dictionary<string, XmlNode>> _indexedSoughtAfterNodes = new Dictionary<int, Dictionary<string, XmlNode>>();
+
+		/// <summary>
+		/// Stores nodes from the xml object that have child nodes which need to be identified by both their key attribute and name
+		/// </summary>
+		private readonly Dictionary<int, Dictionary<Tuple<string,string>, XmlNode>> _indexedSoughtAfterNodes = new Dictionary<int, Dictionary<Tuple<string,string>, XmlNode>>();
 #endif
 
 		public FindByKeyAttribute(string keyAttribute)
@@ -101,18 +105,18 @@ namespace Chorus.merge.xml.generic
 			{
 				_parentsToSearchIn.Add(parentToSearchIn);
 				parentIdx = _parentsToSearchIn.IndexOf(parentToSearchIn);
-				var childrenWithKeys = new Dictionary<string, XmlNode>(); // StringComparer.OrdinalIgnoreCase NO: Bad idea, since I (RBR) saw a case in a data file that had both upper and lower-cased variations.
+				// The child node we want is identified by a combination of it's key attribute and name.
+				var childrenWithKeys = new Dictionary<Tuple<string,string>, XmlNode>(); // StringComparer.OrdinalIgnoreCase NO: Bad idea, since I (RBR) saw a case in a data file that had both upper and lower-cased variations.
 				_indexedSoughtAfterNodes.Add(parentIdx, childrenWithKeys);
-				var matchingName = nodeToMatch.Name;
 				var childrenWithKeyAttr = (from XmlNode childNode in parentToSearchIn.ChildNodes
-										   where childNode.Name == matchingName && childNode.Attributes[_keyAttribute] != null
+										   where childNode.Attributes != null && childNode.Attributes[_keyAttribute] != null
 										   select childNode).ToList();
 				foreach (var nodeWithKeyAttribute in childrenWithKeyAttr)
 				{
 					try
 					{
 
-						childrenWithKeys.Add(nodeWithKeyAttribute.Attributes[_keyAttribute].Value, nodeWithKeyAttribute);
+						childrenWithKeys.Add(new Tuple<string, string>(nodeWithKeyAttribute.Name, nodeWithKeyAttribute.Attributes[_keyAttribute].Value), nodeWithKeyAttribute);
 					}
 					catch(ArgumentException)
 					{
@@ -133,7 +137,7 @@ namespace Chorus.merge.xml.generic
 			}
 
 			XmlNode matchingNode;
-			_indexedSoughtAfterNodes[parentIdx].TryGetValue(key, out matchingNode);
+			_indexedSoughtAfterNodes[parentIdx].TryGetValue(new Tuple<string, string>(nodeToMatch.Name, key), out matchingNode);
 			// JohnT: consider replacing the line above with this if we decide to deal with duplicate keys. This branch currently won't cope with this,
 			// because the Add above will fail on any duplicate key.
 			//if (_indexedSoughtAfterNodes[parentIdx].TryGetValue(key, out matchingNode) && !acceptableTargets.Contains(matchingNode))


sillsdevgerrit:
Can one of the admins verify this patch?

sillsdevgerrit:
Can one of the admins verify this patch?

regnrand:
Review status: 0 of 1 files reviewed at latest revision, 1 unresolved discussion.

---

*[src/LibChorus/merge/xml/generic/FindNodeToMerge.cs, line 108 at r1](https://reviewable.io:443/reviews/sillsdev/chorus/154#-L-seyUbeU9Sx4Bo8Ja0:-L-seyUceSEw1prb_Fym:bdwkw0s) ([raw file](https://github.com/sillsdev/chorus/blob/e5356968497c697e0a4f349320e10d200dc6b732/src/LibChorus/merge/xml/generic/FindNodeToMerge.cs#L108)):*
> ```Smalltalk
> 				var matchingName = nodeToMatch.Name;
> 				var childrenWithKeyAttr = (from XmlNode childNode in parentToSearchIn.ChildNodes
> 										   where childNode.Name == matchingName && childNode.Attributes[_keyAttribute] != null
> ```

Without matching the element name (e.g.: 'matchingName'), it seems like it will just look for any child nodes that have the key attribute, even if that attribute is in some other element. That is probably not desirable. Maybe use "SelectNodes(matchingName)" instead of that over productive "ChildNodes" property.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/sillsdev/chorus/154)*

jasonleenaylor:
*[src/LibChorus/merge/xml/generic/FindNodeToMerge.cs, line 108 at r1](https://reviewable.io:443/reviews/sillsdev/chorus/154#-L-seyUbeU9Sx4Bo8Ja0:-L05FZ_iNejQ5NCIlCZv:b-t26ebc) ([raw file](https://github.com/sillsdev/chorus/blob/e5356968497c697e0a4f349320e10d200dc6b732/src/LibChorus/merge/xml/generic/FindNodeToMerge.cs#L108)):*
<details><summary><i>Previously, regnrand (Randy Regnier) wrote…</i></summary><blockquote>

Without matching the element name (e.g.: 'matchingName'), it seems like it will just look for any child nodes that have the key attribute, even if that attribute is in some other element. That is probably not desirable. Maybe use "SelectNodes(matchingName)" instead of that over productive "ChildNodes" property.
</blockquote></details>

A more complicated fix for this is certainly possible. For context the bug that is being addressed here is caused by the fact that the name of the first nodeToMatch candidate is used and all children matching that are cached. However, if more than one element name is allowed for the children, as found in the features file, all matching for children who have a name different from the first node fails.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/sillsdev/chorus/154)*

regnrand:
Review status: 0 of 1 files reviewed at latest revision, 1 unresolved discussion.

---

*[src/LibChorus/merge/xml/generic/FindNodeToMerge.cs, line 108 at r1](https://reviewable.io:443/reviews/sillsdev/chorus/154#-L-seyUbeU9Sx4Bo8Ja0:-L05OQ5s_W9EALsasMh5:b-vurvff) ([raw file](https://github.com/sillsdev/chorus/blob/e5356968497c697e0a4f349320e10d200dc6b732/src/LibChorus/merge/xml/generic/FindNodeToMerge.cs#L108)):*

A more complicated fix for this is certainly possible. For context the bug that is being addressed here is caused by the fact that the name of the first nodeToMatch candidate is used and all children matching that are cached. However, if more than one element name is allowed for the children, as found in the features file, all matching for children who have a name different from the first node fails.
</blockquote></details>

I may be missing something, but the new comment suggests the bug being fixed is to skip whitespace nodes, which 'ChildNodes' returns (along with comment nodes and actual element nodes). I'm worried about other xml that may have the key attr, but which have different element names which will now be returned. Are you sure that there are none of those in any xml?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/sillsdev/chorus/154)*

RaymondLuong3:
Review status: 0 of 1 files reviewed at latest revision, 1 unresolved discussion.

---

*[src/LibChorus/merge/xml/generic/FindNodeToMerge.cs, line 108 at r1](https://reviewable.io:443/reviews/sillsdev/chorus/154#-L-seyUbeU9Sx4Bo8Ja0:-L05TOJUbjvdHAzTbuPz:bikzhpl) ([raw file](https://github.com/sillsdev/chorus/blob/e5356968497c697e0a4f349320e10d200dc6b732/src/LibChorus/merge/xml/generic/FindNodeToMerge.cs#L108)):*
<details><summary><i>Previously, regnrand (Randy Regnier) wrote…</i></summary><blockquote>

I may be missing something, but the new comment suggests the bug being fixed is to skip whitespace nodes, which 'ChildNodes' returns (along with comment nodes and actual element nodes). I'm worried about other xml that may have the key attr, but which have different element names which will now be returned. Are you sure that there are none of those in any xml?
</blockquote></details>

That is the case. We want to have all ChildNodes returned that have the key attr (even if the element names are different). The bug with duplicating inflection features occurred because children nodes were excluded from being cached because not all children nodes share the same name. Sorry that the comment was misleading. I can fix that too.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/sillsdev/chorus/154)*

regnrand:
Review status: 0 of 1 files reviewed at latest revision, 1 unresolved discussion.

---

*[src/LibChorus/merge/xml/generic/FindNodeToMerge.cs, line 108 at r1](https://reviewable.io:443/reviews/sillsdev/chorus/154#-L-seyUbeU9Sx4Bo8Ja0:-L05V-eFeIHfjj8tE0Yd:b-s19piz) ([raw file](https://github.com/sillsdev/chorus/blob/e5356968497c697e0a4f349320e10d200dc6b732/src/LibChorus/merge/xml/generic/FindNodeToMerge.cs#L108)):*
<details><summary><i>Previously, RaymondLuong3 (Raymond Luong) wrote…</i></summary><blockquote>

That is the case. We want to have all ChildNodes returned that have the key attr (even if the element names are different). The bug with duplicating inflection features occurred because children nodes were excluded from being cached because not all children nodes share the same name. Sorry that the comment was misleading. I can fix that too.
</blockquote></details>

Just be prepared to deal with any number of other merge bugs for non-feature elements, where the old behavior of requiring both element names and key attribute matches is no longer being used.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/sillsdev/chorus/154)*

jasonleenaylor:
*[src/LibChorus/merge/xml/generic/FindNodeToMerge.cs, line 108 at r1](https://reviewable.io:443/reviews/sillsdev/chorus/154#-L-seyUbeU9Sx4Bo8Ja0:-L05_ebPpBuBDzr-quvC:b-q0y1xn) ([raw file](https://github.com/sillsdev/chorus/blob/e5356968497c697e0a4f349320e10d200dc6b732/src/LibChorus/merge/xml/generic/FindNodeToMerge.cs#L108)):*
<details><summary><i>Previously, regnrand (Randy Regnier) wrote…</i></summary><blockquote>

Just be prepared to deal with any number of other merge bugs for non-feature elements, where the old behavior of requiring both element names and key attribute matches is no longer being used.
</blockquote></details>

@raymondluong3 you may want to run dotCover on the unit tests to make sure that this method is not used in other parts of FlexBridge where the element name is important.  Ping me on Slack if you want some help.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/sillsdev/chorus/154)*

papeh:
:lgtm:

---

Reviewed 1 of 1 files at r2.
Review status: all files reviewed at latest revision, 1 unresolved discussion.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/sillsdev/chorus/154#-:-L0GT3rIabl1F7wtMQWU:bnfp4nl)*

PR #156, opened 2018-01-11
Adjust fixutf8 setting in Linux

marksvc:
Remove relative path since a relative path didn't work anyway.
Comment out fixutf8 setting so there is not a potentially
problematic setting left in the way.
Leave commented-out fixutf8 setting, and with an absolute
path, as a template to be corrected manually when needed.

Part of LT-18796.


ermshiperete:
:lgtm: (I assume you made the changes on Windows)

---

Reviewed 2 of 2 files at r1.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/sillsdev/chorus/156#-:-L2dN75dZQb9efdqJxvQ:bknp9v9)*

marksvc:
I did not make the change in the Windows zip. Both because FB in Windows does not have a problem with a bad Mercurial.ini file (because it re-writes it), and IIRC because the setting in the Windows Mercurial.ini looked like a potentially useful default anyway. This came at the expense of consistency; and I can make it consistent if desired.

PR #160, opened 2018-02-06
Always write timestamps with time zones

rmunn:
This is PR #125, but merging into master to make sure the bug stays fixed in future versions of Chorus.

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/chorus/160)

### DIFF
--- a/src/LibChorus/FileTypeHandlers/lift/LiftUtils.cs
+++ b/src/LibChorus/FileTypeHandlers/lift/LiftUtils.cs
@@ -6,9 +6,7 @@ namespace Chorus.FileTypeHandlers.lift
 {
 	public static class LiftUtils
 	{
-		[Obsolete("Use LiftTimeFormatWithTimeZone instead, as LiftTimeFormatNoTimeZone produces incorrect results when used with DateTime.Now")]
-		static public string LiftTimeFormatNoTimeZone = "yyyy-MM-ddTHH:mm:ssZ";
-		static public string LiftTimeFormatWithTimeZone = "yyyy-MM-ddTHH:mm:ssK";
+		public static string LiftTimeFormatWithTimeZone = "yyyy-MM-ddTHH:mm:ssK";
 
 		public static string GetId(XmlNode e)
 		{
--- a/src/LibChorus/FileTypeHandlers/lift/LiftUtils.cs
+++ b/src/LibChorus/FileTypeHandlers/lift/LiftUtils.cs
@@ -6,7 +6,9 @@ namespace Chorus.FileTypeHandlers.lift
 {
 	public static class LiftUtils
 	{
+		[Obsolete("Use LiftTimeFormatWithTimeZone instead, as LiftTimeFormatNoTimeZone produces incorrect results when used with DateTime.Now")]
 		static public string LiftTimeFormatNoTimeZone = "yyyy-MM-ddTHH:mm:ssZ";
+		static public string LiftTimeFormatWithTimeZone = "yyyy-MM-ddTHH:mm:ssK";
 
 		public static string GetId(XmlNode e)
 		{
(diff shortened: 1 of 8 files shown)


ermshiperete:
:lgtm:

---

Reviewed 3 of 8 files at r2, 5 of 5 files at r3.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/sillsdev/chorus/160#-:-L4eFAwcU9rFa7Af9g16:bnfp4nl)*

src/LibChorus/FileTypeHandlers/lift/LiftUtils.cs
```diff
@@ -6,7 +6,9 @@ namespace Chorus.FileTypeHandlers.lift
 {
 	public static class LiftUtils
 	{
+		[Obsolete("Use LiftTimeFormatWithTimeZone instead, as LiftTimeFormatNoTimeZone produces incorrect results when used with DateTime.Now")]
 		static public string LiftTimeFormatNoTimeZone = "yyyy-MM-ddTHH:mm:ssZ";
+		static public string LiftTimeFormatWithTimeZone = "yyyy-MM-ddTHH:mm:ssK";
```
jasonleenaylor: I know this was public, but it looks like there are no examples of use in our github organization. So I would prefer to keep the code clean and just rename without adding the Obsolete comment. And since noone is using it yet we may as well make it internal

src/LibChorus/merge/xml/generic/ChorusNotesMergeEventListener.cs
```diff
@@ -18,7 +18,9 @@ public class ChorusNotesMergeEventListener : IMergeEventListener, IDisposable
 		private FileStream _readerStream;
 		private TempFile _tempFile;
 		private string _path;
+		[Obsolete("Use TimeFormatWithTimeZone instead, as TimeFormatNoTimeZone produces incorrect results when used with DateTime.Now")]
 		static public string TimeFormatNoTimeZone = "yyyy-MM-ddTHH:mm:ssZ";
+		static public string TimeFormatWithTimeZone = "yyyy-MM-ddTHH:mm:ssK";
```
jasonleenaylor: This also is unused outside of Chorus, and while we are at it we could unify all of these into one place.

rmunn: I've been looking through the code and haven't found a totally obvious place to unify these. Put them in LiftUtils? Then it looks strange for the Conflict code to be using `LiftUtils.TimeFormatWithTimeZone`. Make a new class called TimeUtils? Then the patch becomes larger and the chance of conflicts when I cherry-pick it into other branches becomes larger. I'm trying to make a minimal bugfix patch; any refactoring I'd prefer to do in a separate PR after the bugfix goes through.

src/LibChorus/FileTypeHandlers/lift/LiftUtils.cs
rmunn: I'll certainly just clean up the `LiftTimeFormatNoTimeZone` and related constants rather than make them obsolete. OTOH, I'm not a huge fan of making this internal: what do we gain by hiding the time format string? We don't gain the ability to change it: changing it would be a breaking change whether or not it's part of the public-facing API, because it would change the format of the XML produced by Chorus. And since this is used from other namespaces (e.g., XmlUtilities.AddDateCreatedAttribute uses LiftUtils.LiftTimeFormatWithTimeZone), if we were ever to split the Lift and XML-handling parts of Chorus into two separate DLLs, making LiftTimeFormatWithTimeZone internal would just complicate that process.

If you feel that this really *should* be internal I'll go ahead and make that change, but I'll leave it public for now since I believe that's the way cause the fewest possible problems for ourselves in the future.

jasonleenaylor:
I am willing to accept this change now. I didn't realize this was still waiting. Here is one other thing worth considering: It seems to me that it could actually be better to always store the date as UTC in the file. This would make it easier to investigate timing things at the xml level.

PR #161, opened 2018-02-06
Add property to get GUID of the status message

ermshiperete:
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/chorus/161)

### DIFF
--- a/src/Chorus.Tests/UI/Sync/SyncControlModelTests.cs
+++ b/src/Chorus.Tests/UI/Sync/SyncControlModelTests.cs
@@ -22,6 +22,13 @@ namespace Chorus.Tests
 		private ProjectFolderConfiguration _project;
 		private Synchronizer _synchronizer;
 
+		private void WaitForTasksToFinish(int timeoutMinutes, params WaitHandle[] waitHandles)
+		{
+			var timeout = new TimeSpan(0, timeoutMinutes, 0);
+			Assert.That(WaitHandle.WaitAll(waitHandles, (int)timeout.TotalMilliseconds), Is.True,
+				string.Format("Tasks timed out after {0} min.", timeout.TotalMinutes));
+		}
+
 		[SetUp]
 		public void Setup()
 		{
@@ -96,18 +103,10 @@ namespace Chorus.Tests
 			var progress = new ConsoleProgress() {ShowVerbose = true};
 			_model.AddMessagesDisplay(progress);
 			SyncResults results = null;
-			_model.SynchronizeOver += new EventHandler((sender, e) => results = (sender as SyncResults));
+			var waitHandle = new AutoResetEvent(false);
+			_model.SynchronizeOver += (sender, e) => results = (sender as SyncResults);
 			_model.Sync(true);
-			var start = DateTime.Now;
-			while (results == null)
-			{
-				Thread.Sleep(100);
-				Application.DoEvents(); //else the background worker may starve
-				if ((DateTime.Now.Subtract(start).Minutes > 1))
-				{
-					Assert.Fail("Gave up waiting.");
-				}
-			}
+			WaitForTasksToFinish(2, waitHandle);
 			Assert.IsFalse(results.Succeeded);
 			Assert.IsNotNull(results.ErrorEncountered);
 		}
@@ -121,10 +120,11 @@ namespace Chorus.Tests
 			var progress = new ConsoleProgress();
 			_model.AddMessagesDisplay(progress);
 			SyncResults results = null;
-			_model.SynchronizeOver += new EventHandler((sender, e) => results = (sender as SyncResults));
+			_model.SynchronizeOver += (sender, e) => results = sender as SyncResults;
 			_model.Sync(true);
 			Thread.Sleep(100);
 			_model.Cancel();
+			// NOTE: we can't use AutoResetEvent in this test - for some reason this doesn't work
 			var start = DateTime.Now;
 			while (results == null)
 			{
@@ -144,17 +144,9 @@ namespace Chorus.Tests
 		public void AsyncLocalCheckIn_GivesGoodResult()
 		{
 			SyncResults result=null;
-			_model.AsyncLocalCheckIn("testing", (r)=>result=r);
-			var start = DateTime.Now;
-			while (result == null)
-			{
-				Thread.Sleep(100);
-				Application.DoEvents();//without this, the background worker starves 'cause their's no UI
-				if ((DateTime.Now.Subtract(start).Minutes > 0))
-				{
-					Assert.Fail("Gave up waiting.");
-				}
-			}
+			var waitHandle = new AutoResetEvent(false);
+			_model.AsyncLocalCheckIn("testing", r => { result=r; waitHandle.Set();});
+			WaitForTasksToFinish(1, waitHandle);
 			Assert.IsTrue(result.Succeeded);
 		}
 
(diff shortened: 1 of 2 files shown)


papeh:
:lgtm: although it may be slightly safer to `LastMessage()?.GetAttributeValue("guid") ?? string.Empty`

---

Reviewed 1 of 1 files at r1.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/sillsdev/chorus/161#-:-L4fzIQZfNtDSm0Nvg20:b-6o37pu)*

ermshiperete:
I agree, but unfortunately this branch still tracks the chorus-2.5 branch which still uses Mono 3.x, so we can't  use C# 6 features.

---

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/sillsdev/chorus/161#-:-L4g0cKYZOPQyNv4iBat:b-bjdja4)*

rmunn:
Jenkins build is failing on just one test, Chorus.Tests.SyncControlModelTests.AsyncLocalCheckIn_NoHgLockWarningWithMultipleWorkers. The test runs three copies of SyncControlModel.AsyncLocalCheckIn as background workers, and waits one minute before giving up. There's a previous test that runs a single SyncControlModel.AsyncLocalCheckIn, and in the Jenkins build log, that test takes 22 seconds. If the three background workers are competing for resources on a resource-starved system, it's likely that this test would take about 66 seconds to run, so this is a spurious failure that would be solved by bumping the timeout up to, say, 3 minutes:

```diff
diff --git a/src/Chorus.Tests/UI/Sync/SyncControlModelTests.cs b/src/Chorus.Tests/UI/Sync/SyncControlModelTests.cs
index 988fc3b..2f63646 100644
--- a/src/Chorus.Tests/UI/Sync/SyncControlModelTests.cs
+++ b/src/Chorus.Tests/UI/Sync/SyncControlModelTests.cs
@@ -174,7 +174,7 @@ public void AsyncLocalCheckIn_NoHgLockWarningWithMultipleWorkers()
                        {
                                Thread.Sleep(100);
                                Application.DoEvents();//without this, the background worker starves 'cause their's no UI
-                               if(DateTime.Now.Subtract(start).Minutes > 0)
+                               if(DateTime.Now.Subtract(start).Minutes > 2)
                                {
                                        Assert.Fail("Gave up waiting.");
                                }
```

ermshiperete:
Thanks Robin for analysing the test failure! I did some refactoring but basically increased the timeout to 3 minutes for this test.

---

Review status: 1 of 2 files reviewed at latest revision, all discussions resolved.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/sillsdev/chorus/161#-:-L4ki78M7xNRPNgrknN3:b-k53vrw)*

papeh:
:lgtm:

---

Reviewed 1 of 1 files at r2.

---

*[src/Chorus.Tests/UI/Sync/SyncControlModelTests.cs, line 124 at r2](https://reviewable.io:443/reviews/sillsdev/chorus/161#-L4l0eugb09T8Sc-VQGK:-L4l0eugb09T8Sc-VQGL:b-gf89h7) ([raw file](https://github.com/sillsdev/chorus/blob/6f79fd744f1f0ecf8042727a39dcf93e8d34f55a/src/Chorus.Tests/UI/Sync/SyncControlModelTests.cs#L124)):*
> ```Smalltalk
> 			SyncResults results = null;
> 			var waitHandle = new AutoResetEvent(false);
> 			_model.SynchronizeOver += (sender, e) => results = (sender as SyncResults);
> ```

no need to `waitHandle.Set()` for the first two tests?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/sillsdev/chorus/161#-:-L4l0WIvi1tFmeGuIwzc:bnfp4nl)*

ermshiperete:
Review status: 1 of 2 files reviewed at latest revision, 1 unresolved discussion, some commit checks failed.

---

*[src/Chorus.Tests/UI/Sync/SyncControlModelTests.cs, line 124 at r2](https://reviewable.io:443/reviews/sillsdev/chorus/161#-L4l0eugb09T8Sc-VQGK:-L4lWsYU5z5dp_P1n-gc:b4a1sbz) ([raw file](https://github.com/sillsdev/chorus/blob/6f79fd744f1f0ecf8042727a39dcf93e8d34f55a/src/Chorus.Tests/UI/Sync/SyncControlModelTests.cs#L124)):*

no need to `waitHandle.Set()` for the first two tests?
</blockquote></details>

No wonder the test fails :smile:  Thanks! But it turns out that even with correctly setting the waitHandle it still doesn't work  - I guess in this case it needs the Application.DoEvents(). So I reverted back to the previous code for this test.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/sillsdev/chorus/161)*

papeh:
Review status: 1 of 2 files reviewed at latest revision, 2 unresolved discussions, some commit checks failed.

---

*[src/Chorus.Tests/UI/Sync/SyncControlModelTests.cs, line 124 at r2](https://reviewable.io:443/reviews/sillsdev/chorus/161#-L4l0eugb09T8Sc-VQGK:-L4lvUyG2wdKpUQ8zmy0:bcogw3s) ([raw file](https://github.com/sillsdev/chorus/blob/6f79fd744f1f0ecf8042727a39dcf93e8d34f55a/src/Chorus.Tests/UI/Sync/SyncControlModelTests.cs#L124)):*
<details><summary><i>Previously, ermshiperete (Eberhard Beilharz) wrote…</i></summary><blockquote>

No wonder the test fails :smile:  Thanks! But it turns out that even with correctly setting the waitHandle it still doesn't work  - I guess in this case it needs the Application.DoEvents(). So I reverted back to the previous code for this test.
</blockquote></details>

Could you try calling Application.DoEvents(); before calling WaitForTasksToFinish?

---

*[src/Chorus.Tests/UI/Sync/SyncControlModelTests.cs, line 107 at r3](https://reviewable.io:443/reviews/sillsdev/chorus/161#-L4ltbFlcCLJrrpjXUwX:-L4ltbFlcCLJrrpjXUwY:b-unyh1y) ([raw file](https://github.com/sillsdev/chorus/blob/0c0319c2e97e7905295e0b07a30f23c10919a852/src/Chorus.Tests/UI/Sync/SyncControlModelTests.cs#L107)):*
> ```Smalltalk
> 			SyncResults results = null;
> 			var waitHandle = new AutoResetEvent(false);
> 			_model.SynchronizeOver += (sender, e) => results = (sender as SyncResults);
> ```

no need to call `waitHandle.Set()` here?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/sillsdev/chorus/161)*

ermshiperete:
I realized that we don't need the `lfmerge` branch anymore and should work on the `chorus-2.6` branch instead, so I'm closing this PR and creating a new one.

---

Review status: 1 of 2 files reviewed at latest revision, 2 unresolved discussions, some commit checks failed.

---

*[src/Chorus.Tests/UI/Sync/SyncControlModelTests.cs, line 124 at r2](https://reviewable.io:443/reviews/sillsdev/chorus/161#-L4l0eugb09T8Sc-VQGK:-L4oVQ8C9R2MSFWVPwwk:bxmmtfm) ([raw file](https://github.com/sillsdev/chorus/blob/6f79fd744f1f0ecf8042727a39dcf93e8d34f55a/src/Chorus.Tests/UI/Sync/SyncControlModelTests.cs#L124)):*

Could you try calling Application.DoEvents(); before calling WaitForTasksToFinish?
</blockquote></details>

Good idea. Alas, it doesn't work.

---

*[src/Chorus.Tests/UI/Sync/SyncControlModelTests.cs, line 107 at r3](https://reviewable.io:443/reviews/sillsdev/chorus/161#-L4ltbFlcCLJrrpjXUwX:-L4ogbzviKvy4QAZhJS6:brfjp9v) ([raw file](https://github.com/sillsdev/chorus/blob/0c0319c2e97e7905295e0b07a30f23c10919a852/src/Chorus.Tests/UI/Sync/SyncControlModelTests.cs#L107)):*

no need to call `waitHandle.Set()` here?
</blockquote></details>

Thanks for a careful review! The culprit is the Sync() method that doesn't work without plenty of Application.DoEvents(), so I'll had to revert to the while loop for this test.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/sillsdev/chorus/161#-:-L4ogzAxEJN5dDWnnYCW:bt1p4s4)*

ermshiperete:
Actually, we can't use the chorus-2.6 branch because that uses libpalaso 3.1 which we can't use in lfmerge yet. So I'm reopening this PR.

PR #176, opened 2018-06-15
while doing Send/Receive using the new 2.6 Chorus from a repo that wa…

bobeaton:
…s last sync'd with 2.5, we occasionally get an exception, "Failed to set up extensions for the repository: Unable to remove the file to be replaced" (an exception throw from line 174 of HgRepository.cs). Jason suggests changing File.Replace to RobustFile.Replace in the Save method of IniDocument to solve this.

Also, OneStory Editor has ChorusFileType extension DLL that get loaded at run-time and 2.6 has a new behavior that requires the extension to define the class attribute:
[Export(typeof(IChorusFileTypeHandler))] in the FileType handler. Before I knew this was needed, it failed to load the extension DLL and the error was being swallowed, so I added a try...catch around where it fails, so it can throw a more meaningful error message if this wasn't done.

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/chorus/176)

### DIFF
--- a/src/LibChorus/FileTypeHandlers/ChorusFileTypeHandlerCollection.cs
+++ b/src/LibChorus/FileTypeHandlers/ChorusFileTypeHandlerCollection.cs
@@ -50,8 +50,7 @@ namespace Chorus.FileTypeHandlers
                     }
                     catch (ReflectionTypeLoadException ex)
                     {
-                        var typeLoadException = ex as ReflectionTypeLoadException;
-                        var loaderExceptions = typeLoadException.LoaderExceptions;
+                        var loaderExceptions = ex.LoaderExceptions;
                         System.Diagnostics.Debug.Fail("Loading exception!");
                         throw new AggregateException(ex.Message, loaderExceptions);
                     }
--- a/src/LibChorus/FileTypeHandlers/ChorusFileTypeHandlerCollection.cs
+++ b/src/LibChorus/FileTypeHandlers/ChorusFileTypeHandlerCollection.cs
@@ -1,4 +1,4 @@
-﻿// Copyright (c) 2015 SIL International
+// Copyright (c) 2015 SIL International
 // This software is licensed under the MIT License (http://opensource.org/licenses/MIT)
 using System;
 using System.Collections.Generic;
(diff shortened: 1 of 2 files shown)


src/LibChorus/FileTypeHandlers/ChorusFileTypeHandlerCollection.cs
```diff
@@ -44,8 +44,21 @@ public class ChorusFileTypeHandlerCollection
...
 				{
-					container.ComposeParts(this);
-				}
+                    try
+                    {
+                        container.ComposeParts(this);
+                    }
+                    catch (Exception ex)
+                    {
+                        if (ex is ReflectionTypeLoadException)
```
ermshiperete: Wouldn't it be better to directly catch the `ReflectionTypeLoadException` instead of all exceptions?

bobeaton:
There was a change in the way external implementers of the IChorusFileTypeHandler where handled. Starting around 2.5 I think, the new way of incorporating them required the implementers to include an Attribute on their class implementing IChorusFileTypeHandler. But without that attribute, Chorus was failing silently with no explanation. So without wanting to interfere (in case there was a reason this should fail silently), I only meant to put in exception handling for that one case.

But yes, it would have been appropriate to just catch the one. (I might have been thinking to throw all exceptions and chickened out later).

Bob

bobeaton:
> One tiny comment, otherwise :lgtm:
> 
> > Reviewed 2 of 2 files at r1, 2 of 2 files at r2.
> > _[Reviewable](https://reviewable.io/reviews/sillsdev/chorus/176#-:-LcVHWHX29xld4ttOjDb:b-yglm96)_ status: all files reviewed, 2 unresolved discussions (waiting on @bobeaton)
> 
> _[src/LibChorus/FileTypeHandlers/ChorusFileTypeHandlerCollection.cs, line 53 at r2](https://reviewable.io/reviews/sillsdev/chorus/176#-LcVH3qH9K9W6eQR9ByU:-LcVH3qH9K9W6eQR9ByV:bb3eapw) ([raw file](https://github.com/sillsdev/chorus/blob/b3d8f6999ebf42cca4d1bc8cddde8b971f22af01/src/LibChorus/FileTypeHandlers/ChorusFileTypeHandlerCollection.cs#L53)):_
> 
> > ```smalltalk
> >                     catch (ReflectionTypeLoadException ex)
> >                     {
> >                         var typeLoadException = ex as ReflectionTypeLoadException;
> > ```
> 
> No need to cast to `ReflectionTypeLoadException` anymore

sorry, didn't get an email about this.. just checked in the new change removing this.

Let me know when this is merge,

Thanks,Bob

PR #199, opened 2019-11-01
Update NuGet.exe URL

papeh:
TeamCity builds were failing to download NuGet.exe.
Update to the same URL used by libPalaso.

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/chorus/199)


ermshiperete:
I changed the build config on TC to use Mono 6. With that it builds successful, however we ran into a problem when running two tests which I already fixed on my fork in the `feature/nuget` branch. I'd recommend to merge this change and then cherry-pick that commit. It's one of the cases where you wonder why the tests used to pass...

PR #201, opened 2020-01-14
Use SIL.Chorus.Mercurial nuget package

ermshiperete:
This change removes the `MercurialExtensions` and the zipped `Mercurial` files from the git repo and instead includes the `SIL.Chorus.Mercurial` nuget package that contains those files. This has the advantage that dependent projects like flexbridge don't have to download the Mercurial files but instead can include the nuget package.

This change is [Reviewable](https://reviewable.io/reviews/sillsdev/chorus/201)

### DIFF
--- a/lgtm.yml
+++ b/lgtm.yml
@@ -1,0 +1,6 @@
+extraction:
+  # We don't use python in Chorus, so we ignore all files to prevent failing LGTM builds
+  python:
+    index:
+      filters:
+       - exclude: "**/*.*"


src/LibChorus/LibChorus.csproj
```diff
@@ -36,6 +36,7 @@ See full changelog at https://github.com/sillsdev/chorus/blob/feature/nuget/CHAN
     <PackageReference Include="Icu4c.Win.Min" Version="59.1.7" />
     <PackageReference Include="L10NSharp" Version="4.0.2" />
     <PackageReference Include="NDesk.DBus" Version="0.15.0" />
+    <PackageReference Include="SIL.Chorus.Mercurial" Version="3.0.1-beta2" />
```
rmunn: There's a problem with version 3.0.1-beta2 of the NuGet package, which will probably fail to work properly on 32-bit Linux systems; see https://github.com/sillsdev/Mercurial4Chorus/pull/1 for details.

ermshiperete: done

ermshiperete:
Sorry, I accidentally closed it this PR with a wrong push to GitHub.

Issue #191, opened 2019-09-26
Update included Mercurial


ermshiperete:
We're using an old version of Mercurial. We should update this to a current version.

ermshiperete:
Mercurial 3.7 introduces a breaking change (cf https://github.com/sillsdev/chorus/pull/195#issuecomment-544425012) which might cause problems when not all users use the same version of Chorus. A first, safe step would be to update the Linux version to the same version that Windows uses (3.3). It might also be possible to update Windows and Linux to a higher version that doesn't introduce breaking changes (i.e. < 3.7). Or ship both old and new version and allow the client to select which version to use.

rmunn:
I was able to find a build of Mercurial 3.4 at https://launchpad.net/ubuntu/+source/mercurial/3.4-1 which had a 64-bit .deb file still available, so I created #204 with that Mercurial build. It had no 32-bit builds, so the question is now whether we still need 32-bit builds of the Linux version of Mercurial.

If we can abandon 32-bit builds on Linux, then #204 should be good enough to close this. If we still need 32-bit builds, then I need to learn how to cross-compile a 32-bit Python .so module from a .c source, and then I'll be able to produce a 32-bit version of the Mercurial 3.4 package. (@ermshiperete, feel free to tackle that task if you know how to do it!)

papeh:
Ubuntu is no longer producing 32-bit OS's. The FLEx team is hoping, pending analysis of our new analytics, to be able to drop 32-bit support on Windows, too.

marksvc:
Linux FieldWorks 9.1 is/can-be 64-bit only. 
Linux FieldWorks 9.0 was released as 64-bit and 32-bit. It can change to be 64-bit only.

ermshiperete:
While PR #204 tried to update the included Mercurial, things moved along since it got created. We no longer include Mercurial in .zip files but use the [SIL.Chorus.Mercurial](https://github.com/sillsdev/Mercurial4Chorus) nuget package instead, so that needs to be updated.

rmunn:
Four years after opening this issue, I have a viable PR for it in https://github.com/sillsdev/chorus/pull/329 (though https://github.com/sillsdev/Mercurial4Chorus/pull/11 will need to be merged first).

PR #204, opened 2020-02-11
Update included Mercurial to 3.4 for 64-bit Linux

rmunn:
Mercurial 3.4 was the version available in Ubuntu 15.10 (Wily). No 32-bit build of this exists on the web, so we'll eventually have to build that ourselves: the only difference between the 64-bit build and the 32-bit build is six Python modules that Mercurial builds from a C source.

Rationale: #191

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/chorus/204)


rmunn:
Note that the version in the Windows zip file is 3.3, so it's possible there may be some breaking changes between 3.3 and 3.4 that cause us problems. Looking at https://www.mercurial-scm.org/wiki/WhatsNew/Archive#Mercurial_3.4_.282015-05-01.29 for changes marked "(BC)", I don't think anything will cause us issues. I'd like to know what "commands.push: abort when revisions evaluate to empty set (BC)" means in detail (will LfMerge think Mercurial has failed if there was simply no change to push?), but that's what testing on the QA server is for. Most importantly, Mercurial 3.4 is *after* SNI was implemented, but *before* the breaking change in 3.7 that changed how it handled deleted-here-and-changed-over-there files. So in theory, it should still pass all of our unit tests.

ermshiperete:
We no longer ship the .zip files but instead rely on the SIL.Chorus.Mercurial nuget package, so that has to be updated.

PR #206, opened 2020-03-03
Fix LT-20017: Sanitize HtmlDetails

jasonleenaylor:
* Drop any invalid UTF8 sequences before writing the
  HtmlDetails of a conflict to ChorusNotes
* Fix some misspellings

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/chorus/206)

### DIFF
--- a/src/LibChorus/merge/xml/generic/Conflict.cs
+++ b/src/LibChorus/merge/xml/generic/Conflict.cs
@@ -29,7 +29,11 @@ namespace Chorus.merge.xml.generic
 		[Obsolete("Use TimeFormatWithTimeZone instead, as TimeFormatNoTimeZone produces incorrect results when used with DateTime.Now")]
 		public static string TimeFormatNoTimeZone = @"yyyy-MM-ddTHH:mm:ssZ";
 		public static string TimeFormatWithTimeZone = @"yyyy-MM-ddTHH:mm:ssK";
-
+		private static readonly Encoding Utf8Encoder = Encoding.GetEncoding(
+			"UTF-8",
+			new EncoderReplacementFallback(string.Empty), // Drop any problem characters
+			new DecoderExceptionFallback()
+		);
 		private ContextDescriptor _context = new NullContextDescriptor();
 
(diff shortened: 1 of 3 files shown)


jasonleenaylor:
___
*[src/LibChorus/merge/xml/generic/Conflict.cs, line 5 at r1](https://reviewable.io/reviews/sillsdev/chorus/206#-M1XU-AX0uk7XdSrqQuV:-M1XUnQZ86yLrpTbZZnh:byu631c) ([raw file](https://github.com/sillsdev/chorus/blob/046a6831e665478200d515a55c2f1559242859bf/src/LibChorus/merge/xml/generic/Conflict.cs#L5)):*
<details><summary><i>Previously, gmartin7 (Gordon Martin) wrote…</i></summary><blockquote>

Is this for Encoding?
</blockquote></details>

No, it's actually an unused artifact from an earlier fix. I'm removing it and will push up an amendment.

src/LibChorus/merge/xml/generic/Conflict.cs
```diff
@@ -29,7 +29,11 @@ public abstract class Conflict : IConflict, IEquatable<Conflict>
 		[Obsolete("Use TimeFormatWithTimeZone instead, as TimeFormatNoTimeZone produces incorrect results when used with DateTime.Now")]
 		public static string TimeFormatNoTimeZone = @"yyyy-MM-ddTHH:mm:ssZ";
 		public static string TimeFormatWithTimeZone = @"yyyy-MM-ddTHH:mm:ssK";
-
+		private static readonly Encoding Utf8Encoder = Encoding.GetEncoding(
+			"UTF-8",
+			new EncoderReplacementFallback(string.Empty), // Drop any problem characters
```
ermshiperete: I wonder if it would be better to output something like "?" or "(error)" instead of silently hiding the problem by removing the problem characters?
A very fancy way would be to implement our own `EncoderFallback` and `EncoderFallbackBuffer` classes that would output the code point of the problem character.

PR #214, opened 2020-05-22
Use SIL.Chorus.Mercurial nuget package

ermshiperete:
This is the successor of #201 which got accidentally closed.

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/chorus/214)


src/Installer/GeneratedMercurial.wxs
```diff
@@ -2,690 +2,140 @@
 <Wix xmlns="http://schemas.microsoft.com/wix/2006/wi">
     <Fragment>
         <DirectoryRef Id="mercurial">
-            <Component Id="mercurial.add_path.exe" Guid="D63A8C10-90B6-4843-9BCB-187366FEFED0">
+            <Component Id="mercurial.add_path.exe" Guid="29CADE0A-0849-4583-B4D2-524654235A7D">
```
tombogle, line 5: Do we want these GUIds to change?

ermshiperete: I don't know - I know nothing about WIX. I guess @jasonleenaylor or @papeh should take a look.

tombogle: Typically you don't want the GUIDs to change for files that were in previous versions of the installer and are still in the current version. It might just be a matter of efficiency, so if all the files are changing, it might not matter. Usually when we have a wxs file with generated GUIDs, the generator is designed to leave existing ones alone.

ermshiperete: Ah, I see. We have `.guidsForInstaller.xml` files for `MercurialExtensions`, but not for `Mercurial` itself (same thing on the `master` branch) and so we get new GUIDs - unless those files get created on the machine that did it last.

ermshiperete: I wrote a little tool (in SIL.BuildTasks) that can extract the GUIDs from an existing .wxs file and re-create the `.guidsForInstaller.xml` files. I added those files to the `SIL.Chorus.Mercurial` package, and now the GUIDs stay the same :smile:

tombogle:
I notice the LGTM analyses are failing. Not sure if that's significant.

ermshiperete:
> I notice the LGTM analyses are failing. Not sure if that's significant.

No, that can be ignored. I haven't looked why it can't build or how it could be configured.

PR #215, opened 2020-05-22
Update dependencies

ermshiperete:
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/chorus/215)


ermshiperete:
This PR fails to build because it depends on #214, but that shouldn't prevent a review.

PR #237, opened 2021-01-27
Fix generating the configuration with the path for cacert.pem

papeh:
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/chorus/237)


papeh (commented):
Ironically, S&R works fine without this file. It fails only when the file exists and contains a bad path. Should we remove it entirely? :thinking: 

> *[Reviewable](https://reviewable.io/reviews/sillsdev/chorus/237#-:-MS4n1UU20KV8gncaFO0:b3tcuqv)* status: 0 of 1 files reviewed, all discussions resolved

papeh:
oops, pushed the fix to the wrong branch; see #240

PR #244, opened 2021-02-10
[API] Reintroduce default GetCloneFromInternetModel ctor

papeh:
A client was using it. This undoes one of several
+semver:breaking changes from the previous commit.

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/chorus/244)

### DIFF
--- a/src/Chorus/UI/Clone/GetCloneFromInternetModel.cs
+++ b/src/Chorus/UI/Clone/GetCloneFromInternetModel.cs
@@ -12,6 +12,11 @@ namespace Chorus.UI.Clone
 {
 	public class GetCloneFromInternetModel : InternetCloneSettingsModel
 	{
+		public GetCloneFromInternetModel() : base()
+		{
+			ShowCloneOnlyControls = false;
+		}
+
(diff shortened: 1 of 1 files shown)


papeh (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/chorus/244)* status: 0 of 1 files reviewed, 1 unresolved discussion (waiting on @papeh)

___
*[src/Chorus/UI/Clone/GetCloneFromInternetModel.cs, line 25 at r1](https://reviewable.io/reviews/sillsdev/chorus/244#-MTDJHe782JIm22XFQpG:-MTDJHe782JIm22XFQpH:b-f359vm) ([raw file](https://github.com/sillsdev/chorus/blob/b08d0740057019ef5a7a6d3aeefbb5525a58335e/src/Chorus/UI/Clone/GetCloneFromInternetModel.cs#L25)):*
> ```Smalltalk
> 		}
> 
> 		public bool ShowCloneOnlyControls { get; }
> ```

On the other hand, perhaps we should let the removal stand, possibly add a default value of `null` to the other constructor, and remove this equally-unused `ShowCloneOnlyControls` property (and the unused default constructor of the base class)

PR #246, opened 2021-03-31
LT-20555: Bad message for bad username

papeh:
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/chorus/246)

PR #248, opened 2021-04-05
LT-20555: Bad message for bad username (#246)

papeh:
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/chorus/248)


papeh:
included in the next PR

PR #251, opened 2021-04-13
Fix Linux

papeh:
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/chorus/251)


jasonleenaylor (commented):
> Reviewed 5 of 5 files at r1.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/chorus/251)* status: all files reviewed, 1 unresolved discussion (waiting on @papeh)

___
*[src/Chorus/UI/Misc/ServerSettingsControl.Designer.cs, line 151 at r1](https://reviewable.io/reviews/sillsdev/chorus/251#-MYCEtBQ4Bys2vUnnoue:-MYCEtBQ4Bys2vUnnouf:brh52rf) ([raw file](https://github.com/sillsdev/chorus/blob/d77f1901ab24c676a62830dcd2372142e35a622b/src/Chorus/UI/Misc/ServerSettingsControl.Designer.cs#L151)):*
Can you use PlatformUtils.IsMono or something like that?
<details><summary><i alt="summary">Quoted 5 lines of code…</i></summary>

> ```
> #if MONO
> 				120F
> #else
> 				80F
> #endif
> ```
</details>

marksvc (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/chorus/251)* status: all files reviewed, 1 unresolved discussion (waiting on @jasonleenaylor)

___
*[src/Chorus/UI/Misc/ServerSettingsControl.Designer.cs, line 151 at r1](https://reviewable.io/reviews/sillsdev/chorus/251#-MYCEtBQ4Bys2vUnnoue:-MYCJ6bs6MY3zSiP116-:bhgq9gj) ([raw file](https://github.com/sillsdev/chorus/blob/d77f1901ab24c676a62830dcd2372142e35a622b/src/Chorus/UI/Misc/ServerSettingsControl.Designer.cs#L151)):*

Can you use PlatformUtils.IsMono or something like that?

> ```
> #if MONO
> 				120F
> #else
> 				80F
> #endif
> ```

</blockquote></details>

In mono 5 I don't think MONO/similar are defined. They can be manually defined when building, but the new way is to use a run-time check.

README.md
```diff
@@ -146,3 +146,7 @@ build using `UpdateDependencies.sh`:
 	cd chorus
 	./UpdateDependencies.sh
 
+##### Python
+
+Beginning with Ubuntu 20.04 (Focal), you will need to `sudo apt install python-is-python2` so that Mercurial can run.
```
ermshiperete, line 151: I'm not sure that's a wise thing to require since I imagine this will break a lot of other apps that are updated to run with Python 3. Wouldn't it be better to patch our Mercurial package to explicitly use python2 (`/usr/bin/python2` instead of `/usr/bin/python`)?

papeh (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/chorus/251)* status: all files reviewed, 1 unresolved discussion (waiting on @ermshiperete)

___
*[README.md, line 151 at r1](https://reviewable.io/reviews/sillsdev/chorus/251#-MYCDHiE_3AfKXs5kKAa-r1-151:-MYFdXWP4kLUNEc15ORZ:b-ifyv0q) ([raw file](https://github.com/sillsdev/chorus/blob/d77f1901ab24c676a62830dcd2372142e35a622b/README.md#L151)):*
<details><summary><i>Previously, ermshiperete (Eberhard Beilharz) wrote…</i></summary><blockquote>

I'm not sure that's a wise thing to require since I imagine this will break a lot of other apps that are updated to run with Python 3. Wouldn't it be better to patch our Mercurial package to explicitly use python2 (`/usr/bin/python2` instead of `/usr/bin/python`)?
</blockquote></details>

Good idea. Since FLEx Bridge has already been installing py-is-py2, I recommend we let this slide this sprint but try to patch Mercurial before stable.

ermshiperete (approved):
:lgtm:

PR #264, opened 2021-09-10
Use CrossPlatformSettingsProvider

papeh:
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/chorus/264)


papeh:
Clicking Ignore on assert dialogs causes tests to pass locally. If these builds fail, the unit tests are valid.

papeh:
dup of #265

PR #279, opened 2022-04-25
Added .NET Standard target to LibChorus and .NET6 to ChorusMerge.

josephmyers:
Also referencing the appropriate libpalaso .NET Standard DLL's, which drop support for USB operations (RepositoryAddress.cs). For the .NET Standard version, those operations now return null. These methods were already returning null in other situations, but if this is undesirable, we can throw a NotSupportedException, etc.

Also upgraded Autofac version to support the new framework and resolving that reference implicitly in Chorus.csproj.

Also added GH Actions build (and potentially removing TC and AppVeyor builds)

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/chorus/279)


josephmyers:
The new GH workflow, though flaky, successfully builds, tests, and deploys. Deploying only happens for the .NET Standard projects currently, but this can be expanded. The overall goal here is to replace the TC builds.

jasonleenaylor (commented):
> Reviewed 7 of 8 files at r1.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/chorus/279)* status: 7 of 8 files reviewed, 2 unresolved discussions (waiting on @ermshiperete and @josephmyers)

___
*[`.github/workflows/dotnet.yml` line 29 at r1](https://reviewable.io/reviews/sillsdev/chorus/279#-N7fRQL13JGwqZNJSR9M:-N7fRQL13JGwqZNJSR9N:b-qzyihz) ([raw file](https://github.com/sillsdev/chorus/blob/0d7cc1cbe3ef208ca56243541133ee8aa86a9941/.github/workflows/dotnet.yml#L29)):*
> ```YAML
>     - name: Pack
>       if: github.ref == 'refs/heads/develop'
>       run: dotnet pack --no-restore -c Release
> ```

add --no-build?



_Code quote:_
```YAML
un: dotnet pack --no-restore -c Release
```
___
*[`.github/workflows/dotnet.yml` line 35 at r1](https://reviewable.io/reviews/sillsdev/chorus/279#-N7g9zhj4v3WLjnmHNpl:-N7g9zhj4v3WLjnmHNpm:b-y3nuaq) ([raw file](https://github.com/sillsdev/chorus/blob/0d7cc1cbe3ef208ca56243541133ee8aa86a9941/.github/workflows/dotnet.yml#L35)):*
> ```YAML
>     - name: Deploy 2
>       if: github.ref == 'refs/heads/develop'
>       run: dotnet nuget push output\SIL.Chorus.LibChorus.5**\*.nupkg -s https://api.nuget.org/v3/index.json -k ${{secrets.SILLSDEV_PUBLISH_NUGET_ORG}} --skip-duplicate
> ```

I'm not sure why there are two deploys here, and why one is specific to version 5 packages? Seems like that will require extra maintenance.



_Code quote:_
```YAML
    - name: Deploy 1
      if: github.ref == 'refs/heads/develop'
      run: dotnet nuget push output\SIL.Chorus.ChorusMerge**\*.nupkg -s https://api.nuget.org/v3/index.json -k ${{secrets.SILLSDEV_PUBLISH_NUGET_ORG}} --skip-duplicate
    - name: Deploy 2
      if: github.ref == 'refs/heads/develop'
      run: dotnet nuget push output\SIL.Chorus.LibChorus.5**\*.nupkg -s https://api.nuget.org/v3/index.json -k ${{secrets.SILLSDEV_PUBLISH_NUGET_ORG}} --skip-duplicate
```
___
*[`src/Chorus/Chorus.csproj` line 11 at r1](https://reviewable.io/reviews/sillsdev/chorus/279#-N7QogO8EO3EOxBIun0u:-N7QogO8EO3EOxBIun0v:bhstyul) ([raw file](https://github.com/sillsdev/chorus/blob/4600f3979ff5403f20f94795cef7bb77249b1ba6/src/Chorus/Chorus.csproj#L11)):*
> ```XML
> 
>   <ItemGroup>
>     <PackageReference Include="Autofac" Version="2.6.3.862" />
> ```

Ah, so this will come from the project reference now.



_Code quote:_
```XML
    <PackageReference Include="Autofac" Version="2.6.3.862" />
```

jasonleenaylor:
*[`src/ChorusHubApp/ChorusHubApp.csproj` line 14 at r1](https://reviewable.io/reviews/sillsdev/chorus/279#-N7pXfnu5Kab-TzvsYrI:-N7pXfnu5Kab-TzvsYrJ:b32vac) ([raw file](https://github.com/sillsdev/chorus/blob/0d7cc1cbe3ef208ca56243541133ee8aa86a9941/src/ChorusHubApp/ChorusHubApp.csproj#L14)):*
> ```XML
>     <PackageReference Include="Microsoft.SourceLink.GitHub" Version="1.0.0" PrivateAssets="All" />
>     <PackageReference Include="SIL.ReleaseTasks" Version="2.5.0" PrivateAssets="All" />
>     <PackageReference Include="SIL.Windows.Forms" Version="9.0.0-*" />
> ```

10



_Code quote:_
```XML
9
```

papeh (commented):
> Reviewed 1 of 8 files at r1.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/chorus/279)* status: 7 of 8 files reviewed, 3 unresolved discussions (waiting on @ermshiperete and @josephmyers)

___
*[`src/Chorus/Chorus.csproj` line 11 at r1](https://reviewable.io/reviews/sillsdev/chorus/279#-N7QogO8EO3EOxBIun0u:-N7pnabr5kJsA5Pj71nf:bnjzbbv) ([raw file](https://github.com/sillsdev/chorus/blob/4600f3979ff5403f20f94795cef7bb77249b1ba6/src/Chorus/Chorus.csproj#L11)):*

Ah, so this will come from the project reference now.
</blockquote></details>

:tada:

josephmyers:
*[`.github/workflows/dotnet.yml` line 29 at r1](https://reviewable.io/reviews/sillsdev/chorus/279#-N7fRQL13JGwqZNJSR9M:-N7sukbG4V2XGO8TES9O:b-9z5xkl) ([raw file](https://github.com/sillsdev/chorus/blob/0d7cc1cbe3ef208ca56243541133ee8aa86a9941/.github/workflows/dotnet.yml#L29)):*

add --no-build?
</blockquote></details>

sure, sounds good

josephmyers:
*[`.github/workflows/dotnet.yml` line 35 at r1](https://reviewable.io/reviews/sillsdev/chorus/279#-N7g9zhj4v3WLjnmHNpl:-N7sv-YN-DSq62sYgRJa:b-ik3fak) ([raw file](https://github.com/sillsdev/chorus/blob/0d7cc1cbe3ef208ca56243541133ee8aa86a9941/.github/workflows/dotnet.yml#L35)):*

I'm not sure why there are two deploys here, and why one is specific to version 5 packages? Seems like that will require extra maintenance.
</blockquote></details>

good catch. at time of writing i only wanted to push specific builds. we can change this to be more generic, i think "output\**\*.nupkg" would do it and only require one line

josephmyers:
*[`src/ChorusHubApp/ChorusHubApp.csproj` line 14 at r1](https://reviewable.io/reviews/sillsdev/chorus/279#-N7pXfnu5Kab-TzvsYrI:-N7svV1A9tHVADXA7zup:b-l4leuv) ([raw file](https://github.com/sillsdev/chorus/blob/0d7cc1cbe3ef208ca56243541133ee8aa86a9941/src/ChorusHubApp/ChorusHubApp.csproj#L14)):*

10
</blockquote></details>

yeah, rebase should fix!

.github/workflows/dotnet.yml
```diff
@@ -0,0 +1,36 @@
+name: Chorus .NET6
+
+on:
+  [push, pull_request, workflow_dispatch]
+
+jobs:
+  build_test_deploy:
+    env:
+      NUGET_AUTH_TOKEN: ${{secrets.GITHUB_TOKEN}}
+    runs-on: windows-latest
```
ermshiperete, line 10: We should also build on Linux

.github/workflows/dotnet.yml
```diff
@@ -0,0 +1,36 @@
...
+        dotnet-version: 6.0.x
+        source-url: https://nuget.pkg.github.com/sillsdev/index.json
+    - name: Restore
+      run: dotnet restore
+    - name: Build
+      run: dotnet build --no-restore -c Release
+    - name: Test
+      run: dotnet test
+    - name: Pack
+      if: github.ref == 'refs/heads/develop'
```
ermshiperete, line 28: Shouldn't this be `refs/heads/master`? (also below)

.github/workflows/dotnet.yml
josephmyers: I wasn't able to get the Linux side to work. the file system was missing mercurial lib's, if i remember

.github/workflows/dotnet.yml
josephmyers: good catch, will fix

.github/workflows/dotnet.yml
ermshiperete: Would installing the SIL.Chorus.Mercurial nuget package help, or `sudo apt install mercurial` ?

ermshiperete: I see SIL.Chorus.Mercurial is already installed. In my local testing I had to install `sudo apt install python2` to get things working.

.github/workflows/dotnet.yml
```diff
@@ -0,0 +1,32 @@
+name: Chorus .NET6
+
+on:
+  [push, pull_request, workflow_dispatch]
+
+jobs:
+  build_test_deploy:
+    env:
+      NUGET_AUTH_TOKEN: ${{secrets.GITHUB_TOKEN}}
```
ermshiperete, line 9: Why do we need this token?

appveyor.yml
```diff
@@ -2,7 +2,7 @@ version: '{build}'
 branches:
```
ermshiperete, line 2: Since we're adding GHA we don't have any benefits of additionally building on AppVeyor and so should remove this file.

ermshiperete (changes requested):
Can we get rid of the warnings as well? (Maybe in a follow-up PR)

.github/workflows/dotnet.yml
rmunn: Depending on the version of SIL.Chorus.Mercurial and/or the version of Ubuntu, you might also need to run `sudo apt install python-is-python2` (that was necessary in the Microsoft .NET SDK images, for example). That one installs `python2` *and* creates a `/usr/bin/python` symlink to `python2`. (But in newer versions of SIL.Chorus.Mercurial, the Mercurial script is updated to look for `/usr/bin/python2` instead of `/usr/bin/python` in which case you can just install the `python2` package and be fine).

.github/workflows/dotnet.yml
josephmyers: It's required during .NET setup. i removed it to verify, and things break

josephmyers:
> Can we get rid of the warnings as well? (Maybe in a follow-up PR)

To do so, we can either change L10NSharp (not planned) or suppress NU1701. Which do you recommend? There is one other warning, one regarding an unrecommended/obsolete call. I'll look into this now

.github/workflows/dotnet.yml
```diff
@@ -0,0 +1,39 @@
...
+    runs-on: ${{ matrix.os }}
+    strategy:
+      matrix:
+        os: [ubuntu-latest, windows-latest]
+    permissions: write-all
+    
+    steps:
+    - uses: actions/checkout@v3
```
ermshiperete, line 17: From a security POV it would be better to specify the commit SHA instead of the tag - a malicious user/repo owner could change the tag but he can't change the sha.

```suggestion
    - uses: actions/checkout@2541b1294d2704b0964813337f33b291d3f8596b # v3.0.2
```

.github/workflows/dotnet.yml
```diff
@@ -0,0 +1,39 @@
...
+      with:
+        fetch-depth: 0 # fetch full history for GitVersion
+    - name: Setup .NET
+      uses: actions/setup-dotnet@v2
```
ermshiperete, line 21:
```suggestion
      uses: actions/setup-dotnet@c0d4ad69d8bd405d234f1c9166d383b7a4f69ed8 # v2.1.0
```

.github/workflows/dotnet.yml
```diff
@@ -0,0 +1,39 @@
...
+
+jobs:
+  build_test_deploy:
+    env:
+      NUGET_AUTH_TOKEN: ${{secrets.GITHUB_TOKEN}}
+    runs-on: ${{ matrix.os }}
+    strategy:
+      matrix:
+        os: [ubuntu-latest, windows-latest]
+    permissions: write-all
```
ermshiperete, line 14: We don't have the `permissions` line in our other workflow scripts.

.github/workflows/dotnet.yml
```diff
@@ -0,0 +1,39 @@
...
+      with:
+        fetch-depth: 0 # fetch full history for GitVersion
+    - name: Setup .NET
+      uses: actions/setup-dotnet@v2
+      with:
+        dotnet-version: 6.0.x
+        source-url: https://nuget.pkg.github.com/sillsdev/index.json
+    - name: Fix Python on Linux (for mercurial)
+      if: matrix.os == 'ubuntu-latest'
+      run: sudo apt install python-is-python2
```
ermshiperete, line 27: LibChorus references SIL.Chorus.Mercurial 3.0.3.4 which has the explicit reference to python2, so IMO it would be better to install the python2 package. While it probably doesn't matter for GHA it might break the system if you try to follow these steps on a normal machine where you have other Python apps because it redirects python to python2, but on modern systems python is expected to be python3.

.github/workflows/dotnet.yml
```diff
@@ -0,0 +1,39 @@
...
+    - name: Restore
+      run: dotnet restore
+    - name: Build
+      run: dotnet build --no-restore -c Release
+    - name: Test
+      run: dotnet test
```
ermshiperete, line 33:
```suggestion
      run: dotnet test --no-build -c Release
```

.github/workflows/dotnet.yml
```diff
@@ -0,0 +1,39 @@
...
+    - name: Pack
+      if: github.ref == 'refs/heads/master'
```
ermshiperete, line 35: I think we always want to _create_ the nuget package so that you have an artifact to test with.

.github/workflows/dotnet.yml
```diff
@@ -0,0 +1,39 @@
...
+      run: dotnet pack --no-restore --no-build -c Release
+    - name: Deploy
+      if: github.ref == 'refs/heads/master'
```
ermshiperete, line 38: We don't want to do the Deploy step on both Windows and Linux, and to be on the save side also add a check `github.event_name == 'push'` (I'm not entirely sure that's needed, but at least it makes it explicit that it should only happen when a new commit gets pushed to master and not on PRs).

.github/workflows/dotnet.yml
```diff
@@ -0,0 +1,39 @@
...
+      run: dotnet nuget push output\*.nupkg -s https://api.nuget.org/v3/index.json -k ${{secrets.SILLSDEV_PUBLISH_NUGET_ORG}} --skip-duplicate
```
ermshiperete, line 39: We should also add a step that archives the artifacts so they can be used for testing. See the ["Upload Artifacts" step e.g. in icu-dotnet](https://github.com/sillsdev/icu-dotnet/blob/master/.github/workflows/CI-CD.yml).

If you want to get fancy you could add some additional steps as found in icu-dotnet workflow: archiving test results and displaying them in a separate workflow (the event file is needed for that), and doing the publish to nuget as separate job.

ermshiperete (changes requested):
Sorry, there's always "one more thing". But we're getting close :smile:. Thanks for keeping up with my requests :smile:

.github/workflows/dotnet.yml
josephmyers: per y'all's comments yesterday, i tried both mercurial and python. both of those are already installed. this line was the only that made a difference!

.github/workflows/dotnet.yml
```diff
@@ -0,0 +1,56 @@
...
+      with:
+        fetch-depth: 0 # fetch full history for GitVersion
+    - name: Setup .NET
+      uses: actions/setup-dotnet@c0d4ad69d8bd405d234f1c9166d383b7a4f69ed8 # v2.1.0
+      with:
+        dotnet-version: 6.0.x
+        source-url: https://nuget.pkg.github.com/sillsdev/index.json
+    - name: Fix Python on Linux (for mercurial)
+      if: matrix.os == 'ubuntu-latest'
+      run: sudo apt install python-is-python2
```
ermshiperete, line 26: You're right that python2 is already installed. I got it to [work without this build step](https://github.com/ermshiperete/chorus/runs/7991946379?check_suite_focus=true).

```suggestion
```

.github/workflows/dotnet.yml
```diff
@@ -0,0 +1,56 @@
...
+    - name: Restore
+      run: dotnet restore
+    - name: Build
+      run: dotnet build --no-restore -c Release
+    - name: Test
+      run: dotnet test
```
ermshiperete, line 32: A comment would be nice why we can't use `--no-build -c Release`.

```suggestion
    - name: Test
      # it would be nice to run tests with `--no-build -c Release` but that causes test
      # failures on Ubuntu where it suddenly can't open the display.
      run: dotnet test
```

ermshiperete (changes requested):

.github/workflows/dotnet.yml
```diff
@@ -0,0 +1,62 @@
...
+    - name: Setup .NET
+      uses: actions/setup-dotnet@c0d4ad69d8bd405d234f1c9166d383b7a4f69ed8 # v2.1.0
+      with:
+        dotnet-version: 6.0.x
+    - name: Fix Python on Linux (for mercurial)
+      if: matrix.os == 'ubuntu-latest'
+      run: |
+        dpkg -l python
+        dpkg -l python2
+        sudo apt install python2
```
ermshiperete, line 30:
```suggestion
```

ermshiperete (changes requested):

ermshiperete (approved):

PR #280, opened 2022-04-28
LT-20999: Move the location of the local Chorus cache

mark-sil:
Move the location of the local Chorus cache from AppData to the
specific FLEX or WeSay project.  The problem we were seeing was when
the cache in AppData was being used by two applications; in this
case FLEX and WeSay. (Note that the same problem would exist if two
local WeSay projects or two local FLEX project were using the same
Chorus Project ID.)
When the cache was more up to date than the local FLEX or WeSay
project we would see the error described in the defect. To prevent
this problem we now store the cache in the local FLEX or WeSay
project folder, along with the other project information. This
results in one cache per application/project, so the cache should
never have more up to date chorus data than the local FLEX or
WeSay project.
https://jira.sil.org/browse/LT-20999

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/chorus/280)

### DIFF
--- a/src/LibChorus/VcsDrivers/Mercurial/HgResumeTransport.cs
+++ b/src/LibChorus/VcsDrivers/Mercurial/HgResumeTransport.cs
@@ -1,4 +1,4 @@
-﻿using System;
+using System;
 using System.Collections.Generic;
 using System.IO;
 using System.Linq;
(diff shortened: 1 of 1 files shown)


papeh (approved):
Outside of the FieldWorks and FLEx Bridge repos, we should include the full JIRA link, since different Chorus clients may use different ticketing systems.

:lgtm:

> Reviewed 1 of 1 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/chorus/280#-:-N0lhwpW8puWIACNt3bx:b-yqdmod)* status: all files reviewed, 1 unresolved discussion (waiting on @jasonleenaylor and @mark-sil)

___
*[`src/LibChorus/VcsDrivers/Mercurial/HgResumeTransport.cs` line 920 at r1](https://reviewable.io/reviews/sillsdev/chorus/280#-N0lhtBq2RJ6_5VAM8ip:-N0lhtBq2RJ6_5VAM8iq:b7bxxqm) ([raw file](https://github.com/sillsdev/chorus/blob/2b9cfef5c47712b49fb062231b4ae90cfebf479f/src/LibChorus/VcsDrivers/Mercurial/HgResumeTransport.cs#L920)):*
> ```Smalltalk
> 			{
> 				string chorusPath = Path.Combine(_repo.PathToRepo, "Chorus");
> 				return Path.Combine(chorusPath, "ChorusStorage");
> ```

this could be a single path.combine

mark-sil (commented):
I added the link at the bottom of the comment.

> *[Reviewable](https://reviewable.io/reviews/sillsdev/chorus/280#-:-N0mzLUFE7raUxOkZ1t9:b-tbcoa5)* status: 0 of 1 files reviewed, 1 unresolved discussion (waiting on @jasonleenaylor and @papeh)

___
*[`src/LibChorus/VcsDrivers/Mercurial/HgResumeTransport.cs` line 920 at r1](https://reviewable.io/reviews/sillsdev/chorus/280#-N0lhtBq2RJ6_5VAM8ip:-N0mzXB07TxWao_SuGs9:b-896fix) ([raw file](https://github.com/sillsdev/chorus/blob/2b9cfef5c47712b49fb062231b4ae90cfebf479f/src/LibChorus/VcsDrivers/Mercurial/HgResumeTransport.cs#L920)):*

this could be a single path.combine
</blockquote></details>

Done.

PR #296, opened 2022-09-09
Add Microsoft.NET.Test.Sdk to test projects

ermshiperete:
According to the docs[^1] it is recommended to add Microsoft.NET.Test.Sdk to all projects so that the tests can be discovered in VS.

Also add NUnit3TestAdapter to all test projects which is now required[^2] and remove NUnit from Tests-ChorusPlugin where it's not used.

[^1]: https://docs.nunit.org/articles/vs-test-adapter/Adapter-Installation.html
[^2]: https://www.nuget.org/packages/NUnit3TestAdapter/4.2.1#readme-body-tab

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/chorus/296)

### DIFF
--- a/.github/workflows/dotnet.yml
+++ b/.github/workflows/dotnet.yml
@@ -25,10 +25,6 @@ jobs:
       matrix:
         os: [ubuntu-latest, windows-latest]
 
-    env:
-      FILTER: '--filter TestCategory!=RequiresUI'
-      if: runner.os == 'Linux'
-
     steps:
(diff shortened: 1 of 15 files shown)


ermshiperete:
This change now causes more tests to be detected, some of these require UI and thus fail.

ermshiperete:
The failing test will likely be fixed by #295.

papeh (approved):

PR #297, opened 2022-09-16
Net6launch

josephmyers:
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/chorus/297)


.github/workflows/dotnet.yml
```diff
@@ -76,7 +70,6 @@ jobs:
     name: "Publish NuGet package"
     runs-on: ubuntu-latest
     needs: build-and-test
-    if: github.event_name == 'push'
```
ermshiperete, line 79: PLEASE don't publish on every successful PR build - this leads to a lot of manual work to later unlist all the packages of the PR that have a higher version number than our regular pre-release packages.

src/LibChorus/Utilities/ExecutionEnvironment.cs
```diff
@@ -46,6 +46,8 @@ internal static string ChorusMergeFilePath()
 		{
 			if (Platform.IsMono)
 			{
+				//todo can we use ChorusMerge.dll directly? instead of chorusmerge
```
ermshiperete, line 49: I think when running with net6.0 we won't come into this branch, so it's only when running with net461. And in that case we have `ChorusMerge.exe`, not `ChorusMerge.dll`. And on Linux when running on Mono you need the prepended `mono `.

ermshiperete (changes requested):

src/LibChorus/Utilities/ExecutionEnvironment.cs
josephmyers: We shouldn't be running with Mono on Linux anymore. Removing Mono was a major part of the reason to upgrade to .NET6. With .NET6 on Linux, this was pointing directly to the .exe instead of chorusmerge. This change points it back to chorusmerge, and I'm trying to test it

ermshiperete: Sure. But remember that we're still building for net461, and there very likely will be applications that will still run it with Mono, so that option still needs to work.

josephmyers: If Platform.IsLinux isn't equivalent, I could also do (Platform.IsLinux || Platform.IsMono)

josephmyers:
I'm not sure what's going with the latest changes to dotnet.yml, but they're causing the windows tests (net461) to hang. Before, the command was `run: dotnet test -- NUnit.TestOutputXml=TestResults`, but now that it's `run: dotnet test --no-build -c Release --filter TestCategory!=SkipOnBuildServer -- NUnit.TestOutputXml=TestResults`, it hangs during a test.

Here's a snippet from a test run without `--no-build`:
```
  Skipped LaunchDialog_SimulatedUsb_USBHasInvalidRepo
  Skipped ShowIt
  Skipped LaunchDialog_CustomAddress
  Skipped LaunchDialog_FullAddress
Warning: Killing Hg Process...

    Syncing...
    Started at 2022-09-20 03:55:53Z
    Local User: bob
    LanguageForge User: 
```

But with it, it hangs after `Skipped LaunchDialog_FullAddress`.

Unfortunately, removing `--no-build` avoids the hang, but now ChorusMerge.Tests fail and LibChorus.Tests are skipped. Ideas, anyone?

ermshiperete:
Do you know which tests hang? Did you try to ignore (for the purpose of tracking down the problem) the new `HgMergeTests` to see if that makes a difference?

josephmyers:
Even with a diagnostic test run, the current test being run isn't printed (lol), but I figured it out and have the fix pushed. It was related to some Encoding call required for .NET6 but breaking for Framework.

The tests take longer to finish, now that we're running LibChorus on .NET6 as well. I think it's worth it, but I'm open to whatever.

src/LibChorusTests/MiscTests.cs
```diff
@@ -11,10 +11,10 @@ public void AllSettingsUseCrossPlatformProvider()
 		{
 			using (var listener = new SystemAssertListener())
 			{
-				// System.Diagnostics.Debug.Listeners.Add(listener);
-				// // ReSharper disable once ObjectCreationAsStatement because the constructor asserts the conditions we're testing.
-				// new Chorus.Properties.Settings();
-				// Assert.That(listener.Messages, Is.Empty);
+				System.Diagnostics.Trace.Listeners.Add(listener);
```
ermshiperete, line 14: :+1:

src/LibChorusTests/VcsDrivers/Mercurial/HgMergeTests.cs
```diff
@@ -14,41 +14,46 @@ namespace LibChorus.Tests.VcsDrivers.Mercurial
...
-				var exception = Assert.Throws<ApplicationException>(() => hgRepo.Merge("unused", heads.First().Number.LocalRevisionNumber));
+				// pull remote changes
+				localRepo.PullFromTarget("remote", remoteRepoPath);
+				var heads = localRepo.GetHeads();
+
+				var exception = Assert.Throws<ApplicationException>(() =>
+					localRepo.Merge("unused", heads.First().Number.LocalRevisionNumber));
 
 				Assert.IsNotNull(exception);
 				Assert.That(exception.Message.Contains("ChorusMerge Error:"));
```
ermshiperete, line 45: I think this can be changed to something like:

```
Assert.That(() => localRepo.Merge("unused", heads.First().Number.LocalRevisionNumber), 
    Throws.Exception.TypeOf<ApplicationException>().With.Property("Message").Contains("ChorusMerge Error:"));
```

src/LibChorus/Model/ServerSettingsModel.cs
```diff
@@ -352,7 +353,15 @@ internal static string EncryptPassword(string encryptMe)
 			{
 				return encryptMe;
 			}
-			var encryptedData = ProtectedData.Protect(Encoding.Unicode.GetBytes(encryptMe), Encoding.Unicode.GetBytes(EntropyValue), DataProtectionScope.CurrentUser);
+
+			if (RuntimeInformation.IsOSPlatform(OSPlatform.Linux) &&
+				!RuntimeInformation.FrameworkDescription.Contains("Framework"))
```
ermshiperete, line 358: This doesn't work when testing on my local machine. When running with Mono `RuntimeInformation.FrameworkDescription` returns `Mono 6.12.0.182 (tarball Tue Jun 14 22:35:00 UTC 2022)`.

src/ChorusMerge/Program.cs
```diff
@@ -41,6 +42,11 @@ public static int Main(string[] args)
 
 				if (Platform.IsWindows)
 				{
+					if (!RuntimeInformation.FrameworkDescription.Contains("Framework"))
```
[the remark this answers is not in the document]
ermshiperete: Answering my own question: there's a nuget package that can be installed that allows to use `RuntimeInformation` in net461.

src/ChorusMerge/ChorusMerge.csproj
```diff
@@ -13,6 +13,7 @@
     <PackageReference Include="GitVersion.MsBuild" Version="5.10.3" PrivateAssets="All" />
     <PackageReference Include="Microsoft.SourceLink.GitHub" Version="1.1.1" PrivateAssets="All" />
     <PackageReference Include="SIL.ReleaseTasks" Version="2.5.0" PrivateAssets="All" />
+    <PackageReference Include="System.Text.Encoding.CodePages" Version="4.0.1" />
```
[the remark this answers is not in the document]
josephmyers: We could, but it would make the Program.cs change need preprocessor directives to compile. I don't think the reference is hurting anything

src/ChorusMerge/Program.cs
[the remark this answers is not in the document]
josephmyers: I've been compiling and testing with Framework. If you're seeing something different, lemme know

src/LibChorus/Model/ServerSettingsModel.cs
josephmyers: Ok, I'm unable to see this on my Linux box. Should I replace Framework with something? Maybe Mono? Do I need another condition? Or I could try deleting it?

josephmyers: What command are you using to run with Mono? It's quite possible I'm not using the right thing

josephmyers: I reverted this code. It's unrelated to the core bug, which was just Platform.IsMono -> .IsLinux, and created [a separate issue for it](https://github.com/sillsdev/chorus/issues/298).

ermshiperete: I was stepping through the net461 tests in the debugger in Rider on Ubuntu 22.04.

ermshiperete (approved):

src/LibChorus/Model/ServerSettingsModel.cs
josephmyers: Ok. Yeah I have less of a pure Linux environment, being on WSL, so I trust your PC more than mine

PR #301, opened 2022-11-09
increase audio and picture file size limit to 10MB

### LINKED ISSUE #277: Increase 1-megabyte file size limit
The [maximum file size](https://github.com/sillsdev/chorus/blob/4600f3979ff5403f20f94795cef7bb77249b1ba6/src/LibChorus/FileTypeHandlers/DefaultFileTypeHandler.cs#L87) allowable in repositories that Chorus handles is 1 megabyte, for reasons [discussed here](https://github.com/sillsdev/chorus/blob/d6e008da344d3d6bf64f9c127e47ee13c4ad0348/src/LibChorus/sync/LargeFileFilter.cs#L7). That commit was made in 2011. Many of those reasons still apply (we don't want to let people shoot their repositories in the foot by adding hour-long video files, for example), but the 1-megabyte limit is becoming too small. Most people have phones with cameras that take 3- or 4-megabyte photos, and if they want to add a photo to their repository, they have to learn how to trim its size and reduce its JPG quality. CPUs and network speeds have also increased, on average, over the past decade, so that larger file sizes, such as 5 or 10 megabytes, probably won't harm repositories all that much now.

So while we still don't want to allow adding video files to Chorus repositories, I believe it's time to consider increasing the file size. A limit of 5 megabytes would allow photos taken by most low- to mid-grade consumer cameras these days, but many high-grade phones are taking photos that are even larger: 6 or 8 megabytes. So I propose increasing the file size limit from 1 megabyte to 10 megabytes. That will allow photos taken by most cameras to be added to a repository. It would also mean that .wav files longer than 5.6 seconds would be possible as well: a .wav file would have to reach 56.69 seconds before its size passed the 10-megabyte limit.

megahirt:
Fixes #277 

This is in response to users who report a need to add audio and pictures to their project over 1MB in FLEx and Language Forge.  The 1MB is reported to be too small.

Because we don't know practically how this will impact the majority of users, I am happy to introduce this change and evaluate the impact (pluses and minuses) on Language Forge and FLEx users who use Language Depot to collaborate, as well as the increased server storage needed to accommodate the higher limits.

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/chorus/301)

### DIFF
--- a/src/LibChorus/FileTypeHandlers/audio/AudioFileTypeHandler.cs
+++ b/src/LibChorus/FileTypeHandlers/audio/AudioFileTypeHandler.cs
@@ -79,7 +79,7 @@ namespace Chorus.FileTypeHandlers.audio
 		/// </remarks>
 		public uint MaximumFileSize
 		{
-			get { return LargeFileFilter.Megabyte; }
+			get { return LargeFileFilter.Megabyte * 10; }
 		}
 	}
 }
--- a/src/LibChorus/FileTypeHandlers/image/ImageFileTypeHandler.cs
+++ b/src/LibChorus/FileTypeHandlers/image/ImageFileTypeHandler.cs
@@ -83,7 +83,7 @@ namespace Chorus.FileTypeHandlers.image
 		/// </remarks>
 		public uint MaximumFileSize
 		{
-			get { return LargeFileFilter.Megabyte; }
+			get { return LargeFileFilter.Megabyte * 10; }
 		}
 	}
 }


papeh (approved):
:lgtm:

> Reviewed 2 of 2 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/chorus/301#-:-NIEq39W7pLB7pyiN0oY:bnfp4nl)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @megahirt)

megahirt:
Hi @papeh, I'm confused about the workflow in this PR.  It says that the PR was merged with failing unit tests (why was it merged?) but then the commit was later reverted "until unit tests can be fixed".  Does this mean that the change is no longer in master?

PR #302, opened 2022-11-23
Improve Get from Internet UX (attempt)

papeh:
* Always show the log in URL (so users know into where they're logging)
* Hide the folder name (and download button) control until the user has logged in or specified Custom URL
* Revert the log in button's text to "&Log in"
* Move the Log in button above the Custom URL control so it's closer to the username and password
* Remove the obsolete TranslationMemory.Tmx

This addresses https://jira.sil.org/browse/LT-21209

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/chorus/302)

### DIFF
--- a/src/Chorus/UI/Misc/ServerSettingsControl.cs
+++ b/src/Chorus/UI/Misc/ServerSettingsControl.cs
@@ -20,11 +20,7 @@ namespace Chorus.UI.Misc
 		{
 			InitializeComponent();
 
-			_serverLabel.Visible = ServerSettingsModel.IsQaServer || ServerSettingsModel.IsPrivateServer;
-			if (_serverLabel.Visible)
-			{
-				_serverLabel.Text = string.Format(_serverLabel.Text, $"{(ServerSettingsModel.IsPrivateServer ? "private" : null)}{ServerSettingsModel.LanguageForgeServer}");
-			}
+			_serverLabel.Text = string.Format(_serverLabel.Text, $"{(ServerSettingsModel.IsPrivateServer ? "private" : "public")}{ServerSettingsModel.LanguageForgeServer}");
 
 			_bandwidth.Items.AddRange(ServerSettingsModel.Bandwidths);
 		}
--- a/src/LibChorus/Properties/launchSettings.json
+++ b/src/LibChorus/Properties/launchSettings.json
@@ -1,9 +1,0 @@
-{
-  "profiles": {
-    "LibChorus": {
-      "commandName": "Executable",
-      "executablePath": "C:\\fwrepo\\flexbridge\\output\\Debug\\net461\\FLExBridge.exe",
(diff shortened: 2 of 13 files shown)


src/Chorus/UI/Misc/ServerSettingsControl.cs
```diff
@@ -20,11 +20,7 @@ public ServerSettingsControl()
 		{
 			InitializeComponent();
 
-			_serverLabel.Visible = ServerSettingsModel.IsQaServer || ServerSettingsModel.IsPrivateServer;
-			if (_serverLabel.Visible)
-			{
-				_serverLabel.Text = string.Format(_serverLabel.Text, $"{(ServerSettingsModel.IsPrivateServer ? "private" : null)}{ServerSettingsModel.LanguageForgeServer}");
-			}
+			_serverLabel.Text = string.Format(_serverLabel.Text, $"{(ServerSettingsModel.IsPrivateServer ? "private" : "public")}{ServerSettingsModel.LanguageForgeServer}");
```
ermshiperete, line 23: I'm surprised we don't need/want a space after `private`/`public`, but that's nothing new...

src/LibChorus/Properties/launchSettings.json
```diff
@@ -0,0 +1,9 @@
+{
+  "profiles": {
+    "LibChorus": {
+      "commandName": "Executable",
+      "executablePath": "C:\\fwrepo\\flexbridge\\output\\Debug\\net461\\FLExBridge.exe",
+      "commandLineArgs": "-u developer -p c:\\fwrepo\\fw\\DistFiles\\Projects -v obtain -projDir c:\\fwrepo\\fw\\DistFiles\\Projects -fwAppsDir C:\\fwrepo\\fw\\Output\\Debug -fwmodel 7000072 -liftmodel 0.13_ldml3 -locale en -pipeID SendReceivec:\\fwrepo\\fw\\DistFiles\\Projectsobtain"
```
ermshiperete, line 6: I'm not sure if it makes sense to check in this file since it contains hard-coded paths. 

I can understand that it might be useful to have this setup available. I'd prefer if we would put it in the readme, or rename this file to something like `launchSettings.json.sample` - that way it doesn't imply that it should work out of the box. And we should put `launchSettings.json` in the `.gitignore` file so that it doesn't get accidentally committed.

papeh: done

src/Chorus/UI/Misc/ServerSettingsControl.cs
papeh: The formatted text is "to public.languageforge.org" or "to private-qa.languageforge.org" or similar.

ermshiperete (approved):

PR #303, opened 2022-11-28
Add Help button to Get from Internet dialog

papeh:
Part of https://jira.sil.org/browse/LT-21209

This change is [Reviewable](https://reviewable.io/reviews/sillsdev/chorus/303)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -19,6 +19,7 @@ Mercurial/
 mercurial/
 *ReSharper*
 build/old
+launchSettings.json
 lib/$(Configuration)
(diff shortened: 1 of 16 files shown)


ermshiperete:
Is this different from #305 (other than that the other contains more changes) ?

papeh:
> Is this different from #305 (other than that the other contains more changes) ?

No. I was trying to create discrete PRs as I completed deliverable chunks of work, but GitHub doesn't handle those the same way Gerrit does.

PR #305, opened 2022-11-29
Clarify that languageforge.org is Language Depot

papeh:
and update localization notes.

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/chorus/305)

### DIFF
--- a/src/Chorus/UI/Misc/ServerSettingsControl.cs
+++ b/src/Chorus/UI/Misc/ServerSettingsControl.cs
@@ -20,11 +20,7 @@ namespace Chorus.UI.Misc
 		{
 			InitializeComponent();
 
-			_serverLabel.Visible = ServerSettingsModel.IsQaServer || ServerSettingsModel.IsPrivateServer;
-			if (_serverLabel.Visible)
-			{
-				_serverLabel.Text = string.Format(_serverLabel.Text, $"{(ServerSettingsModel.IsPrivateServer ? "private" : null)}{ServerSettingsModel.LanguageForgeServer}");
-			}
+			_serverLabel.Text = string.Format(_serverLabel.Text, $"{(ServerSettingsModel.IsPrivateServer ? "private" : "public")}{ServerSettingsModel.LanguageForgeServer}");
 
 			_bandwidth.Items.AddRange(ServerSettingsModel.Bandwidths);
 		}
--- a/src/LibChorus/Model/ServerSettingsModel.cs
+++ b/src/LibChorus/Model/ServerSettingsModel.cs
@@ -79,7 +79,7 @@ namespace Chorus.Model
 
 		static ServerSettingsModel()
 		{
-			Bandwidths = new[] {new BandwidthItem(BandwidthEnum.Low), new BandwidthItem(BandwidthEnum.High)};
+			Bandwidths = new[] {new BandwidthItem(BandwidthEnum.High), new BandwidthItem(BandwidthEnum.Low)};
 		}
 		#endregion static and constant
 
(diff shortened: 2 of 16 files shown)


src/LibChorus/Model/ServerSettingsModel.cs
```diff
@@ -31,6 +30,7 @@ public static string LanguageForgeServer
 			}
```
ermshiperete, line 30: Nothing new, but I find it confusing that `LanguageForgeServer` returns a name with leading dot. Also, we should check that `lfServer` starts with a dot.

src/Chorus/UI/Misc/ServerSettingsControl.cs
```diff
@@ -20,11 +20,7 @@ public ServerSettingsControl()
 		{
 			InitializeComponent();
 
-			_serverLabel.Visible = ServerSettingsModel.IsQaServer || ServerSettingsModel.IsPrivateServer;
-			if (_serverLabel.Visible)
-			{
-				_serverLabel.Text = string.Format(_serverLabel.Text, $"{(ServerSettingsModel.IsPrivateServer ? "private" : null)}{ServerSettingsModel.LanguageForgeServer}");
-			}
+			_serverLabel.Text = string.Format(_serverLabel.Text, $"{(ServerSettingsModel.IsPrivateServer ? "private" : "public")}{ServerSettingsModel.LanguageForgeServer}");
```
ermshiperete, line 23: A comment with the (typical) string we end up with might be helpful.

src/LibChorus/Properties/launchSettings.json.sample
```diff
@@ -0,0 +1,14 @@
...
+      "executablePath": "C:\\fwrepo\\flexbridge\\output\\Debug\\net461\\FLExBridge.exe",
+      "commandLineArgs": "-u developer -p c:\\fwrepo\\fw\\DistFiles\\Projects -v obtain -projDir c:\\fwrepo\\fw\\DistFiles\\Projects -fwAppsDir C:\\fwrepo\\fw\\Output\\Debug -fwmodel 7000072 -liftmodel 0.13_ldml3 -locale en -pipeID SendReceivec:\\fwrepo\\fw\\DistFiles\\Projectsobtain"
+    },
+    "FLEx Bridge (S&R)": {
+      "commandName": "Executable",
+      "executablePath": "C:\\fwrepo\\flexbridge\\output\\Debug\\net461\\FLExBridge.exe",
+      "commandLineArgs": "-u developer -p c:\\fwrepo\\fw\\DistFiles\\Projects\\Sena 3\\Sena 3.fwdata -v send_receive -f C:\\fwrepo\\fw\\Output\\Debug\\FixFwData.exe -projDir c:\\fwrepo\\fw\\DistFiles\\Projects -fwAppsDir C:\\fwrepo\\fw\\Output\\Debug -fwmodel 7000072 -liftmodel 0.13_ldml3 -locale en -pipeID SendReceivec:\\fwrepo\\fw\\DistFiles\\Projects\\Sena 3\\Sena 3.fwdatasend_receive -ws seh"
+    }
+  }
+}
```
ermshiperete, line 14: :+1:

ermshiperete (approved):
:lgtm:

src/Chorus/UI/Misc/ServerSettingsControl.cs
papeh: Would `// format the string with the server URL. We build our own URL because 'resumable' is not helpful here.` suffice?

src/LibChorus/Model/ServerSettingsModel.cs
papeh: I'll add a comment. It needs to include the leading dot so it can be replaced with the leading hyphen in -qa.lf.org

PR #319, opened 2023-06-20
Add webm as audio-file-type so it gets 10MB limit

myieye:
We [recently increased](https://github.com/sillsdev/chorus/pull/301) the max file size for audio and images to 10MB.

But webm is a significant file format in the browser (i.e. Language Forge) and it's not included in the list of known audio file types, so recording audio in LF never actually benefited from this previous change 😲😢,

Webm is a file-type that FLEx supports, so there's no obvious reason to me why we shouldn't include it here, it was simply forgotten.
image

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/chorus/319)

### DIFF
--- a/src/LibChorus/FileTypeHandlers/audio/AudioFileTypeHandler.cs
+++ b/src/LibChorus/FileTypeHandlers/audio/AudioFileTypeHandler.cs
@@ -68,7 +68,7 @@ namespace Chorus.FileTypeHandlers.audio
 		/// <returns>A collection of extensions (without leading period (.)) that can be processed.</returns>
 		public IEnumerable<string> GetExtensionsOfKnownTextFileTypes()
 		{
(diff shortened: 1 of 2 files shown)


ermshiperete (approved):

hahn-kev:
one reason that webm might have not been included is that FLEx doesn't support it in all cases. There's some cases where a webm file is encoded using PCM audio (by the browser) and that encoding isn't supported as a webm file in FLEx. Because of this issue in the Combine we decided to convert webm files to wav when they were exported to FLEx. We might run into the same issue here where files will now get exported properly, and FLEx will see them, but still can't play them.

myieye:
@hahn-kev Interesting. Well, you could already send webm files to FLEx before this change. They just reverted to the default file size limit. All I did was increase the file size limit for them.

PR #325, opened 2023-09-28
Add public StartClone() method

josephmyers:
This is necessary so that flexbridge can supply required cloning information and start a repo download without direct/manual user input. This order will come from Fieldworks when launched from CMD or protocol handler with the proper arguments.

Original issue [here](https://github.com/sillsdev/languageforge-lexbox/issues/242).

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/chorus/325)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,6 +19,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 ### Added
 
 - [SIL.Chorus.LibChorus] Add webm as additional audio file type
+- [SIL.Chorus] Add ability to clone project without direct user interaction
 
 ### Changed
 
--- a/src/Chorus/UI/Clone/GetCloneFromInternetDialog.cs
+++ b/src/Chorus/UI/Clone/GetCloneFromInternetDialog.cs
@@ -262,10 +262,26 @@ namespace Chorus.UI.Clone
 
 		private void OnDownloadClick(object sender, EventArgs e)
 		{
+			StartClone();
+		}
+
+		public void StartClone(string username, string password, Uri targetUri)
+		{
+			_model.Username = username;
+			_model.Password = password;
+			_model.CustomUrl = targetUri.ToString();
+			_model.IsCustomUrl = true;
+			_model.LocalFolderName = GetProjectName(targetUri);
+
+			StartClone();
+		}
+
+		private void StartClone()
+		{
 			lock (this)
 			{
 				_logBox.Clear();
-				if(_backgroundWorker.IsBusy)
+				if (_backgroundWorker.IsBusy)
 					return;
 				UpdateDisplay(State.MakingClone);
 				_model.SaveUserSettings();
@@ -275,6 +291,10 @@ namespace Chorus.UI.Clone
 			}
 		}
 
+		private string GetProjectName(Uri uri)
+		{
+			return uri.Segments[1].ToString();
+		}
 
 		public string ThreadSafeUrl
 		{
--- a/src/Chorus/UI/Clone/GetCloneFromInternetDialog.cs
+++ b/src/Chorus/UI/Clone/GetCloneFromInternetDialog.cs
@@ -270,18 +270,26 @@ namespace Chorus.UI.Clone
 		/// Starts a clone operation with the supplied information.
 		/// <param name="username">Username for Mercurial authentication</param>
 		/// <param name="password">Password for Mercurial authentication</param>
+		/// <param name="projectFolder">The parent directory to put the clone in</param>
 		/// <param name="projectName">Name of the project to clone</param>
 		/// <param name="projectUri">URI where the project can be found</param>
 		/// </summary>
-		public void StartClone(string username, string password, string projectName, Uri projectUri)
+		public static GetCloneFromInternetDialog StartClone(string username, string password, string projectFolder, string projectName, Uri projectUri)
 		{
-			_model.Username = username;
-			_model.Password = password;
-			_model.CustomUrl = projectUri.ToString();
-			_model.IsCustomUrl = true;
-			_model.LocalFolderName = projectName;
+			var model = new GetCloneFromInternetModel(projectFolder)
+			{
+				Username = username,
+				Password = password,
+				CustomUrl = projectUri.ToString(),
+				IsCustomUrl = true,
+				LocalFolderName = projectName
+			};
 
-			StartClone();
+			var dialog = new GetCloneFromInternetDialog(model);
+			dialog.Show();
+			dialog.StartClone();
+
+			return dialog;
 		}
 
 		private void StartClone()
--- a/src/Chorus/UI/Clone/GetCloneFromInternetDialog.cs
+++ b/src/Chorus/UI/Clone/GetCloneFromInternetDialog.cs
@@ -265,13 +265,21 @@ namespace Chorus.UI.Clone
 			StartClone();
 		}
 
-		public void StartClone(string username, string password, Uri targetUri)
+		
+		/// <summary>
+		/// Starts a clone operation with the supplied information.
+		/// <param name="username">Username for Mercurial authentication</param>
+		/// <param name="password">Password for Mercurial authentication</param>
+		/// <param name="host">Host name (with scheme), e.g. https://www.google.com</param>
+		/// <param name="projectName">Name of the project to clone (will be combined with the host Uri)</param>
+		/// </summary>
+		public void StartClone(string username, string password, Uri host, string projectName)
 		{
 			_model.Username = username;
 			_model.Password = password;
-			_model.CustomUrl = targetUri.ToString();
+			_model.CustomUrl = new Uri(host, projectName).ToString();
 			_model.IsCustomUrl = true;
-			_model.LocalFolderName = GetProjectName(targetUri);
+			_model.LocalFolderName = projectName;
 
 			StartClone();
 		}
@@ -289,11 +297,6 @@ namespace Chorus.UI.Clone
 				//_backgroundWorker.RunWorkerAsync(new object[] { ThreadSafeUrl, PathToNewProject, _progress });
 				_backgroundWorker.RunWorkerAsync(new object[0]);
 			}
-		}
-
-		private string GetProjectName(Uri uri)
-		{
-			return uri.Segments[1].ToString();
 		}
 
 		public string ThreadSafeUrl
--- a/src/Chorus/UI/Clone/GetCloneFromInternetDialog.cs
+++ b/src/Chorus/UI/Clone/GetCloneFromInternetDialog.cs
@@ -270,14 +270,14 @@ namespace Chorus.UI.Clone
 		/// Starts a clone operation with the supplied information.
 		/// <param name="username">Username for Mercurial authentication</param>
 		/// <param name="password">Password for Mercurial authentication</param>
-		/// <param name="host">Host name (with scheme), e.g. https://www.google.com</param>
-		/// <param name="projectName">Name of the project to clone (will be combined with the host Uri)</param>
+		/// <param name="projectName">Name of the project to clone</param>
+		/// <param name="projectUri">URI where the project can be found</param>
 		/// </summary>
-		public void StartClone(string username, string password, Uri host, string projectName)
+		public void StartClone(string username, string password, string projectName, Uri projectUri)
 		{
 			_model.Username = username;
 			_model.Password = password;
-			_model.CustomUrl = new Uri(host, projectName).ToString();
+			_model.CustomUrl = projectUri.ToString();
 			_model.IsCustomUrl = true;
 			_model.LocalFolderName = projectName;
 
--- a/src/Chorus/UI/Clone/GetCloneFromInternetDialog.cs
+++ b/src/Chorus/UI/Clone/GetCloneFromInternetDialog.cs
@@ -130,20 +130,24 @@ namespace Chorus.UI.Clone
 		/// <param name="projectUri">URI where the project can be found</param>
 		public static CloneResult DoClone(string username, string password, string projectFolder, string projectName, string projectUri)
 		{
-			var model = new GetCloneFromInternetModel(projectFolder)
-			{
-				Username = username,
-				Password = password,
-				CustomUrl = projectUri,
-				IsCustomUrl = true,
-				LocalFolderName = projectName
-			};
+			var model = new GetCloneFromInternetModel(projectFolder);
+			var oldUser = model.Username;
+			var oldPass = model.Password;
+			model.Username = username;
+			model.Password = password;
+			model.CustomUrl = projectUri;
+			model.IsCustomUrl = true;
+			model.LocalFolderName = projectName;
 
 			using (var dialog = new GetCloneFromInternetDialog(model))
 			{
 				dialog.Show();
 				dialog.StartClone();
 				Application.Run(dialog);
+
+				model.Username = oldUser;
+				model.Password = oldPass;
+				model.SaveUserSettings();
 
 				return GetSharedProjectModel.GetResult(dialog.DialogResult, dialog.PathToNewlyClonedFolder);
 			}
--- a/src/Chorus/UI/Clone/GetCloneFromInternetDialog.cs
+++ b/src/Chorus/UI/Clone/GetCloneFromInternetDialog.cs
@@ -128,7 +128,8 @@ namespace Chorus.UI.Clone
 		/// <param name="projectFolder">The parent directory to put the clone in</param>
 		/// <param name="projectName">Name for the project on the local machine</param>
 		/// <param name="projectUri">URI where the project can be found</param>
-		public static CloneResult DoClone(string username, string password, string projectFolder, string projectName, string projectUri)
+		/// <param name="saveUserSettings">Flag to persist username and password in settings</param>
+		public static CloneResult DoClone(string username, string password, string projectFolder, string projectName, string projectUri, bool saveUserSettings = true)
 		{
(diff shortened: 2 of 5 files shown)


jasonleenaylor (commented):
Since this is an API change that we are interested in I think it deserves a version bump. Adding +semver:major in the commit comment will take care of that.

> Reviewed 1 of 1 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/chorus/325#-:-NfREJy50noyWkDIObBp:b-18d6lu)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @josephmyers)

src/Chorus/UI/Clone/GetCloneFromInternetDialog.cs
```diff
@@ -261,11 +261,27 @@ private void _cancelButton_Click(object sender, EventArgs e)
...
+			StartClone();
+		}
+
+		public void StartClone(string username, string password, Uri targetUri)
+		{
+			_model.Username = username;
+			_model.Password = password;
+			_model.CustomUrl = targetUri.ToString();
+			_model.IsCustomUrl = true;
+			_model.LocalFolderName = GetProjectName(targetUri);
```
hahn-kev, line 274: I'd rather have this passed in explicitly then trying to determine it from the uri, if the uri format changes then this will need to be rewritten. Right now it depends on the project name being somewhere specific in the uri but there's not really anything else that requires that to be the case.

josephmyers: Sure, done.

ermshiperete:
This change deserves a line in `CHANGELOG.md`.

src/Chorus/UI/Clone/GetCloneFromInternetDialog.cs
```diff
@@ -261,11 +261,35 @@ private void _cancelButton_Click(object sender, EventArgs e)
...
+
+		
+		/// <summary>
+		/// Starts a clone operation with the supplied information.
+		/// <param name="username">Username for Mercurial authentication</param>
+		/// <param name="password">Password for Mercurial authentication</param>
+		/// <param name="projectName">Name of the project to clone</param>
+		/// <param name="projectUri">URI where the project can be found</param>
+		/// </summary>
+		public void StartClone(string username, string password, string projectName, Uri projectUri)
```
papeh, line 276: I think the public method would make more sense as a static method to

* create a `GetCloneFromInternetModel` from the arguments
* create a `GetCloneFromInternetDialog` from the model
* StartClone(); with no arguments

papeh:
I'm a bit unclear. The commit comment on https://github.com/sillsdev/flexbridge/pull/388 says that a dialog will be shown to confirm the download, and `GetCloneFromInternetDialog` seems to be the only dialog shown, but in this PR, it looks like StartClone() bypasses all user interaction.

Will there be user interaction? (I assume you wouldn't want the user to edit anything, except perhaps the local project name.)

hahn-kev:
@papeh there will be little to no user interaction. The idea is that a user clicks a button in Language Depot (in the browser) and FLEx opens and clones the project. There is a dialog that FLEx pops up (since it's what gets invoked by the protocol handler) which would prompt the user "Download Sena 3 from LanguageDepot.org?" as a security precaution (any website could invoke the protocol handler and we don't want FLEx to download stuff off the internet automatically without user permission). We could skip the dialog for authorized domains (such as LanguageDepot) but prompt for unexpected domains. I don't want to block unknown domains because it would be useful for testing, or if someone wanted to self host language depot.

src/Chorus/UI/Clone/GetCloneFromInternetDialog.cs
```diff
@@ -261,11 +261,43 @@ private void _cancelButton_Click(object sender, EventArgs e)
...
+		
+		/// <summary>
+		/// Starts a clone operation with the supplied information.
+		/// <param name="username">Username for Mercurial authentication</param>
+		/// <param name="password">Password for Mercurial authentication</param>
+		/// <param name="projectFolder">The parent directory to put the clone in</param>
+		/// <param name="projectName">Name of the project to clone</param>
+		/// <param name="projectUri">URI where the project can be found</param>
+		/// </summary>
+		public static GetCloneFromInternetDialog StartClone(string username, string password, string projectFolder, string projectName, Uri projectUri)
```
papeh, line 277: Nitpicking: This dialog seems kind of a funny place for this method. I'd consider `GetSharedProjectModel` or even a new class.

CHANGELOG.md
```diff
@@ -19,6 +19,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 ### Added
 
 - [SIL.Chorus.LibChorus] Add webm as additional audio file type
+- [SIL.Chorus] Add ability to clone project without direct user interaction
```
papeh, line 22: "without user interaction": there's no indirect interaction, either.

papeh (changes requested):

CHANGELOG.md
josephmyers: In my mind, the indirect user interaction is them using the wrapping software. They still have to do something to get the clone to happen.

src/Chorus/UI/Clone/GetCloneFromInternetDialog.cs
josephmyers: If you'd like me to move it, let me know where. (It doesn't matter too much to me.) I put it here, because it needs the model and dialog, both of which are present here. But its being static now makes separation easier

src/Chorus/UI/Clone/GetCloneFromInternetDialog.cs
```diff
@@ -261,11 +261,43 @@ private void _cancelButton_Click(object sender, EventArgs e)
...
+			StartClone();
+		}
+
+		
+		/// <summary>
+		/// Starts a clone operation with the supplied information.
+		/// <param name="username">Username for Mercurial authentication</param>
+		/// <param name="password">Password for Mercurial authentication</param>
+		/// <param name="projectFolder">The parent directory to put the clone in</param>
+		/// <param name="projectName">Name of the project to clone</param>
```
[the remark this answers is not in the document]
josephmyers: Sure!

josephmyers:
@papeh @jasonleenaylor If there are things you'd like done to these PR's, especially comments that are minor or don't affect the code itself, feel free to do those. I'm not strongly opinionated on the vast majority of this. Doing so would avoid a not less than 24hr communication cycle time. But I also know you guys are super busy, so I'm more than happy to do them as soon as I can. Whatever I can do to help things along. (Sometimes it takes longer to make the comment than to make the change 😄)

josephmyers:
@papeh I did your first two comments and will wait for more guidance on the second two. Let me know what you think. Thanks

papeh (changes requested):
> Reviewed 1 of 2 files at r4, 1 of 1 files at r5, all commit messages.
> Dismissed @hahn-kev and @papeh from 2 discussions.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/chorus/325)* status: all files reviewed, 3 unresolved discussions (waiting on @hahn-kev, @jasonleenaylor, and @josephmyers)

___
*[`src/Chorus/UI/Clone/GetCloneFromInternetDialog.cs` line 276 at r3](https://reviewable.io/reviews/sillsdev/chorus/325#gh-1341572205:-Ng4MPjB1TGhTWsowD9z:bb6un1j) ([raw file](https://github.com/sillsdev/chorus/blob/0f884e83fbe22fa98f6b1a64dd74fded78683f2a/src/Chorus/UI/Clone/GetCloneFromInternetDialog.cs#L276)):*

I think the public method would make more sense as a static method to

* create a `GetCloneFromInternetModel` from the arguments
* create a `GetCloneFromInternetDialog` from the model
* StartClone(); with no arguments
</blockquote></details>

Done
___
*[`src/Chorus/UI/Clone/GetCloneFromInternetDialog.cs` line 127 at r5](https://reviewable.io/reviews/sillsdev/chorus/325#-Ng4NZd_3pL7DDFkPY0V:-Ng4NZd_3pL7DDFkPY0W:be3fl47) ([raw file](https://github.com/sillsdev/chorus/blob/6d98401451651555e9a9f32a9499489fe23ce6aa/src/Chorus/UI/Clone/GetCloneFromInternetDialog.cs#L127)):*
> ```Smalltalk
> 			Application.Run(dialog);
> 
> 			var cloneStatus = res == DialogResult.OK ? CloneStatus.Created : CloneStatus.NotCreated;
> ```

see `GetSharedProjectModel` for more robust CloneStatus parsing

papeh (commented):
> Reviewed 4 of 4 files at r6, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/chorus/325)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @hahn-kev and @jasonleenaylor)

___
*[`src/Chorus/UI/Clone/GetCloneFromInternetDialog.cs` line 277 at r4](https://reviewable.io/reviews/sillsdev/chorus/325#gh-1347668058:-Ng61cvT0Z04zt712Ssa:b-n1r70f) ([raw file](https://github.com/sillsdev/chorus/blob/31df94d0530831cf56a3c32f8a9e816baff7e3be/src/Chorus/UI/Clone/GetCloneFromInternetDialog.cs#L277)):*
<details><summary><i>Previously, josephmyers wrote…</i></summary><blockquote>

If you'd like me to move it, let me know where. (It doesn't matter too much to me.) I put it here, because it needs the model and dialog, both of which are present here. But its being static now makes separation easier
</blockquote></details>

I considered moving it but didn't bother:
* StartClone can stay private (instead of internal)
* DoClone is clearly specific to Internet clones (although that would be apparent from the args, as neither CH nor USB take credentials)

josephmyers:
Looking good to me. Always a fan of making things simpler!

papeh (changes requested):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/chorus/325)* status: all files reviewed, 1 unresolved discussion (waiting on @hahn-kev, @jasonleenaylor, and @josephmyers)

___
*[`src/Chorus/UI/Clone/GetCloneFromInternetDialog.cs` line 306 at r6](https://reviewable.io/reviews/sillsdev/chorus/325#-NgKNtMf8-dzpkmmS9Pj:-NgKNtMf8-dzpkmmS9Pk:bvaw0tm) ([raw file](https://github.com/sillsdev/chorus/blob/57ce43469552845a8a269cc9fcc3159ff5be2bb0/src/Chorus/UI/Clone/GetCloneFromInternetDialog.cs#L306)):*
> ```Smalltalk
> 					return;
> 				UpdateDisplay(State.MakingClone);
> 				_model.SaveUserSettings();
> ```

Is it a problem that the passed `bearer:token` will overwrite the user's saved credentials, if any?

josephmyers:
> > _[Reviewable](https://reviewable.io/reviews/sillsdev/chorus/325)_ status: all files reviewed, 1 unresolved discussion (waiting on @hahn-kev, @jasonleenaylor, and @josephmyers)
> 
> _[`src/Chorus/UI/Clone/GetCloneFromInternetDialog.cs` line 306 at r6](https://reviewable.io/reviews/sillsdev/chorus/325#-NgKNtMf8-dzpkmmS9Pj:-NgKNtMf8-dzpkmmS9Pk:bvaw0tm) ([raw file](https://github.com/sillsdev/chorus/blob/57ce43469552845a8a269cc9fcc3159ff5be2bb0/src/Chorus/UI/Clone/GetCloneFromInternetDialog.cs#L306)):_
> 
> > ```
> > 					return;
> > 				UpdateDisplay(State.MakingClone);
> > 				_model.SaveUserSettings();
> > ```
> 
> Is it a problem that the passed `bearer:token` will overwrite the user's saved credentials, if any?

@hahn-kev

hahn-kev:
*[`src/Chorus/UI/Clone/GetCloneFromInternetDialog.cs` line 306 at r6](https://reviewable.io/reviews/sillsdev/chorus/325#-NgKNtMf8-dzpkmmS9Pj:-NgWmy3wAhcsAf5g3Ksw:bd5fgyk) ([raw file](https://github.com/sillsdev/chorus/blob/57ce43469552845a8a269cc9fcc3159ff5be2bb0/src/Chorus/UI/Clone/GetCloneFromInternetDialog.cs#L306)):*
<details><summary><i>Previously, josephmyers wrote…</i></summary><blockquote>

@hahn-kev 
</blockquote></details>

the token can be used for any project the user has access to, so it should be ok. @jasonleenaylor we can talk about this next time we chat if you want.

josephmyers (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/chorus/325)* status: 4 of 5 files reviewed, 1 unresolved discussion (waiting on @hahn-kev, @jasonleenaylor, and @papeh)

___
*[`src/Chorus/UI/Clone/GetCloneFromInternetDialog.cs` line 306 at r6](https://reviewable.io/reviews/sillsdev/chorus/325#-NgKNtMf8-dzpkmmS9Pj:-NgXxt4d1Z1MTlXj6BK_:b-h2vvy3) ([raw file](https://github.com/sillsdev/chorus/blob/57ce43469552845a8a269cc9fcc3159ff5be2bb0/src/Chorus/UI/Clone/GetCloneFromInternetDialog.cs#L306)):*
<details><summary><i>Previously, hahn-kev (Kevin Hahn) wrote…</i></summary><blockquote>

the token can be used for any project the user has access to, so it should be ok. @jasonleenaylor we can talk about this next time we chat if you want.
</blockquote></details>

Good catch, Hasso. I submitted a PoC hack to demonstrate reverting user/pass, though we'd want some sophistication if we went this route. Kevin also mentioned extending JWT lifecycles and various other tricks you can do with them.

josephmyers (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/chorus/325)* status: 4 of 5 files reviewed, 1 unresolved discussion (waiting on @hahn-kev, @jasonleenaylor, and @papeh)

___
*[`src/Chorus/UI/Clone/GetCloneFromInternetDialog.cs` line 306 at r6](https://reviewable.io/reviews/sillsdev/chorus/325#-NgKNtMf8-dzpkmmS9Pj:-NgYC0rnBqVOJGYE5lJZ:b-1ll3do) ([raw file](https://github.com/sillsdev/chorus/blob/57ce43469552845a8a269cc9fcc3159ff5be2bb0/src/Chorus/UI/Clone/GetCloneFromInternetDialog.cs#L306)):*
<details><summary><i>Previously, josephmyers wrote…</i></summary><blockquote>

Good catch, Hasso. I submitted a PoC hack to demonstrate reverting user/pass, though we'd want some sophistication if we went this route. Kevin also mentioned extending JWT lifecycles and various other tricks you can do with them.
</blockquote></details>

Kevin just mentioned that we would probably only want to revert, if at all, if we have something to revert back to.

papeh (changes requested):
> Reviewed 1 of 1 files at r7, 1 of 1 files at r9, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/chorus/325)* status: all files reviewed, 1 unresolved discussion (waiting on @hahn-kev and @jasonleenaylor)

___
*[`src/Chorus/UI/Clone/GetCloneFromInternetDialog.cs` line 306 at r6](https://reviewable.io/reviews/sillsdev/chorus/325#-NgKNtMf8-dzpkmmS9Pj:-NgZTdbsAU3jY2JSEYoO:b-dlh7ss) ([raw file](https://github.com/sillsdev/chorus/blob/57ce43469552845a8a269cc9fcc3159ff5be2bb0/src/Chorus/UI/Clone/GetCloneFromInternetDialog.cs#L306)):*
<details><summary><i>Previously, josephmyers wrote…</i></summary><blockquote>

Kevin just mentioned that we would probably only want to revert, if at all, if we have something to revert back to.
</blockquote></details>

Thanks for clarifying. Even if the token works, users would be confused if they see 'bearer' instead of their username in the Chorus settings dialog.

josephmyers:
> Thanks for clarifying. Even if the token works, users would be confused if they see 'bearer' instead of their username in the Chorus settings dialog.

True! So I'm fine reverting it all the time, to prevent that. I cleaned it up. Let me know what you think

josephmyers (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/chorus/325)* status: 4 of 5 files reviewed, 1 unresolved discussion (waiting on @hahn-kev, @jasonleenaylor, and @papeh)

___
*[`src/Chorus/UI/Clone/GetCloneFromInternetDialog.cs` line 306 at r6](https://reviewable.io/reviews/sillsdev/chorus/325#-NgKNtMf8-dzpkmmS9Pj:-NgauGnz6a37HdmstDqB:b-ybbtt9) ([raw file](https://github.com/sillsdev/chorus/blob/57ce43469552845a8a269cc9fcc3159ff5be2bb0/src/Chorus/UI/Clone/GetCloneFromInternetDialog.cs#L306)):*

Thanks for clarifying. Even if the token works, users would be confused if they see 'bearer' instead of their username in the Chorus settings dialog.
</blockquote></details>

True! So I'm fine reverting it all the time, to prevent that. I cleaned it up. Let me know what you think

papeh (approved):
:lgtm:

> Reviewed 1 of 1 files at r11, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/chorus/325#-:-NgdROJ15BJSSADKHa-0:bnfp4nl)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @hahn-kev and @jasonleenaylor)

PR #332, opened 2024-03-14
Also build for net8.0

rmunn:
LfMerge wants to move to run on net8.0 rather than net6.0; [this will require a chorusmerge build that can run on net8.0](https://github.com/sillsdev/LfMerge/actions/runs/8284368432/job/22670066938).

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/chorus/332)

### DIFF
--- a/.github/workflows/ci+cd.yml
+++ b/.github/workflows/ci+cd.yml
@@ -52,24 +52,20 @@ jobs:
     # Unfortunately we need two test steps because we need different filters.
     # We could conditionally set an environment variable, but unfortunately
     # the syntax to access that is different on Windows vs Linux.
+    - name: Test on Linux
+      run: dotnet test --no-build -c Release --filter TestCategory!=RequiresUI -- NUnit.TestOutputXml=TestResults
+      if: matrix.os == 'ubuntu-latest'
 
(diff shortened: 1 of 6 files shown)


.github/workflows/ci+cd.yml
```diff
@@ -33,16 +33,18 @@ jobs:
         fetch-depth: 0 # fetch full history for GitVersion
 
     - name: Setup .NET
-      uses: actions/setup-dotnet@c0d4ad69d8bd405d234f1c9166d383b7a4f69ed8 # v2.1.0
+      uses: actions/setup-dotnet@v4.0.0
```
ermshiperete, line 36: For security reasons it's recommended to use the sha1 (which is immutable) instead of the tag (which can be changed in the repo). Probably/hopefully doesn't matter too much here for actions under GitHub's control, but...

ermshiperete (changes requested):
LGTM, except one thing that is more on the nitpick side...

.github/workflows/ci+cd.yml
rmunn: Done in commit 8fe98b56.

ermshiperete (approved):
:LGTM:

PR #336, opened 2024-04-11
switch away from thread abort where we can

### LINKED ISSUE #335: dotnet no longer supports thread abort
if you call Thread abort in any new dotnet code it will throw an exception. This is a problem in Chorus because it's used in the HgProcess output reader here:
https://github.com/sillsdev/chorus/blob/04e1e281629260cdc89e718f418d9c0491859a44/src/LibChorus/Utilities/HgProcessOutputReader.cs#L87-L90

Ultimately the code each of those threads is running needs to be rewritten to check something like a cancellation token which we could cancel there instead of aborting the threads.

hahn-kev:
closes #335

while writing my tests I discovered that HgProcessOutputReader throws exceptions if redirects are not setup, so I added a comment about that and removed all the code designed to support optional redirects.

I also had to change the ThreadState checks as the threads go to sleep now while waiting to read from the stream, from what I can tell in my tests this new version works all the same.

There were 2 process streams, one was unused so I removed it.


jasonleenaylor:
*[`src/ChorusHub/Advertiser.cs` line 72 at r2](https://reviewable.io/reviews/sillsdev/chorus/336#-NvJPnGv5ELko5AXtfPN:-NvJPnGw6O1MvpVTFqWl:b6kq14j) ([raw file](https://github.com/sillsdev/chorus/blob/23507913767e770fe5fe6bd56521800a12a0fb64/src/ChorusHub/Advertiser.cs#L72)):*
> ```Smalltalk
> 			}
> 
> 			if (_cancellationTokenSource.Token.IsCancellationRequested && cancelled)
> ```

Would all conditions that cause a ThreadAbortException also set IsCancellationRequested on the token? Or will it not matter under those circumstances?

hahn-kev:
*[`src/ChorusHub/Advertiser.cs` line 72 at r2](https://reviewable.io/reviews/sillsdev/chorus/336#-NvJPnGv5ELko5AXtfPN:-NvJb9GrCdb4ycf6K7K7:b-92omc0) ([raw file](https://github.com/sillsdev/chorus/blob/23507913767e770fe5fe6bd56521800a12a0fb64/src/ChorusHub/Advertiser.cs#L72)):*

Would all conditions that cause a ThreadAbortException also set IsCancellationRequested on the token? Or will it not matter under those circumstances?
</blockquote></details>

in theory there won't be anymore ThreadAbortExceptions, but if there was then IsCancellationRequested would probably be false. So this should probably be ||

hahn-kev:
*[`src/ChorusHub/Advertiser.cs` line 72 at r2](https://reviewable.io/reviews/sillsdev/chorus/336#-NvJPnGv5ELko5AXtfPN:-NvJcf997ESSpxUQWUR4:bewkt3c) ([raw file](https://github.com/sillsdev/chorus/blob/23507913767e770fe5fe6bd56521800a12a0fb64/src/ChorusHub/Advertiser.cs#L72)):*
<details><summary><i>Previously, hahn-kev (Kevin Hahn) wrote…</i></summary><blockquote>

in theory there won't be anymore ThreadAbortExceptions, but if there was then IsCancellationRequested would probably be false. So this should probably be ||
</blockquote></details>

on the other hand, should we just always close the client? It seems like it should always have been closed and maybe it was just a mistake that it wasn't?

jasonleenaylor:
*[`src/ChorusHub/Advertiser.cs` line 72 at r2](https://reviewable.io/reviews/sillsdev/chorus/336#-NvJPnGv5ELko5AXtfPN:-NvJgK3_3ox_aq85beM9:b-buz88g) ([raw file](https://github.com/sillsdev/chorus/blob/23507913767e770fe5fe6bd56521800a12a0fb64/src/ChorusHub/Advertiser.cs#L72)):*
<details><summary><i>Previously, hahn-kev (Kevin Hahn) wrote…</i></summary><blockquote>

on the other hand, should we just always close the client? It seems like it should always have been closed and maybe it was just a mistake that it wasn't?
</blockquote></details>

I was thinking it should probably be an || for safety, or just remove the ThreadAbortException if it could only have been triggered by our code. It looks like the client was only closed on Stop via the Abort before and stop is called in the Dispose method.

jasonleenaylor (commented):
> Reviewed 6 of 8 files at r1, 2 of 2 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/chorus/336)* status: all files reviewed, 1 unresolved discussion (waiting on @hahn-kev)

hahn-kev:
*[`src/ChorusHub/Advertiser.cs` line 72 at r2](https://reviewable.io/reviews/sillsdev/chorus/336#-NvJPnGv5ELko5AXtfPN:-NvY03lN6jFlzXWWnPok:bcw2z10) ([raw file](https://github.com/sillsdev/chorus/blob/23507913767e770fe5fe6bd56521800a12a0fb64/src/ChorusHub/Advertiser.cs#L72)):*

I was thinking it should probably be an || for safety, or just remove the ThreadAbortException if it could only have been triggered by our code. It looks like the client was only closed on Stop via the Abort before and stop is called in the Dispose method.
</blockquote></details>

alright, I just removed all this and I'm now closing and disposing of the client in Stop. If this class is going to be reused then start needs to be called again, so stop is the logical place to cleanup the client.

jasonleenaylor (approved):
:lgtm:

> Reviewed 1 of 1 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/chorus/336#-:-NvY4giF8jfFh4F1_Umi:bnfp4nl)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @hahn-kev)

PR #343, opened 2024-08-07
fix case where port number is lost using resumable

hahn-kev:
previously the port number would be lost when using resumable. Meaning if the url was typed in as localhost:5158 requests would be made to localhost port 80.

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/chorus/343)

### DIFF
--- a/.github/workflows/ci+cd.yml
+++ b/.github/workflows/ci+cd.yml
@@ -21,10 +21,14 @@ jobs:
 
   build-and-test:
     runs-on: ${{ matrix.os }}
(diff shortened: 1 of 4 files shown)


megahirt:
When i look into the GHA tests for chorus, it appears that the build-and-test on windows-latest was killed due to a 60 minute time limit.  AFAICT all the windows tests passed up until the build was cancelled at 59:59.

megahirt (approved):
LGTM assuming the tests all pass

PR #347, opened 2024-08-14
Don't duplicate hg process output in GetTextFromQuery

### LINKED ISSUE #345: Don't duplicate stdout and stderr in verbose progress output
HgRepository.GetTextFromQuery calls ExecuteErrorsOk, then prints the contents of stdout and stderr via `progress.WriteVerbose` if they aren't empty. However, ExecuteErrorsOk *also* prints stdout and stderr with `progress.WriteVerbose` if they aren't empty. This duplication does nothing but make it harder to find things in the Chorus output, and I suggest getting rid of the `progress.WriteVerbose` calls from HgRepository.GetTextFromQuery.

rmunn:
Fixes #345.

GetTextFromQuery calls ExecuteErrorsOk, which already logs the value of stdout and stderr to the progress dialog. So there's no point in logging the same value a second time.

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/chorus/347)

### DIFF
--- a/src/LibChorus/VcsDrivers/Mercurial/HgRepository.cs
+++ b/src/LibChorus/VcsDrivers/Mercurial/HgRepository.cs
@@ -468,18 +468,6 @@ namespace Chorus.VcsDrivers.Mercurial
 		{
 			var result = ExecuteErrorsOk(query, secondsBeforeTimeoutOnLocalOperation);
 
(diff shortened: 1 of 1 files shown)


rmunn:
PR is on top of #346 in order to avoid merge conflicts once #346 is merged into master.

hahn-kev (approved):
nice catch

PR #349, opened 2024-08-14
No more unnecessary empty "stderr:" lines in log

### LINKED ISSUE #348: Don't write empty "standerr:" lines to verbose progress output
Many times the verbose progress log will contain things like the following:

```
   Using cached result: log -r0 --template "{node}" 
   standerr: 
   
   standout: 4173623d30060fb329fffb3d90fb36027aeb6a83
   
   Using cached result: log -r0 --template "{node}" 
   standerr: 
   
   standout: 4173623d30060fb329fffb3d90fb36027aeb6a83
```

There's no need to write the `standerr` section if it's blank, and there's also no need to double up on the final newline from standard output. `IProgress.WriteVerbose` already writes a newline, so we should trim the extra newline before writing it, and skip writing something entirely if there's no meaningful output (it's just a newline, for example).

rmunn:
Many times stderr's contents are just a single newline; there's no need to write that to the progress log. We also don't need to write double newlines to the log, so trim the final trailing newline before writing.

Fixes #348.

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/chorus/349)

### DIFF
--- a/src/LibChorus/VcsDrivers/Mercurial/HgRepository.cs
+++ b/src/LibChorus/VcsDrivers/Mercurial/HgRepository.cs
@@ -702,13 +702,13 @@ namespace Chorus.VcsDrivers.Mercurial
 			{
 				throw new UserCancelledException();
 			}
-			if (!string.IsNullOrEmpty(result.StandardError))
+			if (!string.IsNullOrWhiteSpace(result.StandardError))
 			{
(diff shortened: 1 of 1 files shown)


rmunn:
PR created on top of #347 to avoid merge conflicts once #347 is merged to master.

hahn-kev (approved):

PR #353, opened 2024-10-29
Allow overriding hg.exe location with env vars

### LINKED ISSUE #352: Allow overriding path to Mercurial via environment variable
In order to test https://github.com/sillsdev/chorus/issues/350 and similar changes more easily, it might be useful to be able to specify the Mercurial folder and/or the name of the binary to run via an environment variable, similarly to how the CHORUSDEBUGGING and CHORUS_LOCAL_TIMEOUT_SECONDS environment variables allow Chorus's behavior to be controlled.

I suggest CHORUS_PATH_TO_MERCURIAL_FOLDER to set the folder without setting the exe name, CHORUS_HG_EXE to set the name of the executable to run without specifying the path (so that one can switch from hg to chg and back), and CHORUS_PATH_TO_HG_EXECUTABLE to set the entire path (if that one is set, then CHORUS_PATH_TO_MERCURIAL_FOLDER and CHORUS_HG_EXE are both ignored). These names line up with the names of the constants in src/LibChorus/MercurialLocation.cs.

The alternative is to have to change the Chorus code (and build a new NuGet package) if we want to use `chg` instead of `hg`.

rmunn:
Fix #352.

Three env vars are defined, one that sets the path to the Mercurial executable directly (overriding both folder and exe name), one that just overrides the folder and lets Chorus pick the exe name, and one that just overrides the exe name and lets Chorus pick the folder.

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/chorus/353)

### DIFF
--- a/src/LibChorus/MercurialLocation.cs
+++ b/src/LibChorus/MercurialLocation.cs
@@ -54,7 +54,7 @@ namespace Chorus// DON'T MOVE THIS! It needs to be super easy for the client to 
 			get
 			{
 				string path = Environment.GetEnvironmentVariable(EnvPathToHgExecutable);
-				if (path != null) return path;
+				if (!string.IsNullOrEmpty(path)) return path;
 				GuessAtLocationIfNotSetAlready();
 
 				if(string.IsNullOrEmpty(_pathToMercurialFolder))
--- a/src/LibChorus/MercurialLocation.cs
+++ b/src/LibChorus/MercurialLocation.cs
@@ -1,4 +1,5 @@
-﻿using System.IO;
+using System;
+using System.IO;
 using Chorus.Utilities;
 using SIL.Code;
 using SIL.PlatformUtilities;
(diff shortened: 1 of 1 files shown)


src/LibChorus/MercurialLocation.cs
```diff
@@ -42,12 +47,14 @@ public static string PathToMercurialFolder
...
 		/// <summary>
-		/// Will use the PathToMercurialFolder, otherwise will just return "hg"
+		/// Will use the environment variable override if set, or will use PathToMercurialFolder if possible, otherwise will just return "hg" and rely on PATH
 		/// </summary>
 		public static string PathToHgExecutable
 		{
 			get
 			{
+				string path = Environment.GetEnvironmentVariable(EnvPathToHgExecutable);
+				if (path != null) return path;
```
hahn-kev, line 57: this should probably use a null or empty check

hahn-kev (changes requested):
looks good, I would make one change so that if the `CHORUS_PATH_TO_HG_EXECUTABLE` variable is an empty string it won't do anything

src/LibChorus/MercurialLocation.cs
rmunn: Fixed in commit 179726ab.

hahn-kev (approved):

PR #354, opened 2024-10-29
Run tests in parallel, one per framework

rmunn:
Now that tests run on three frameworks (net461, net6.0, net8.0), we can save a lot of build time by running each framework's tests in parallel.

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/chorus/354)

### DIFF
--- a/.github/workflows/ci+cd.yml
+++ b/.github/workflows/ci+cd.yml
@@ -61,14 +61,16 @@ jobs:
     # Unfortunately we need two test steps because we need different filters.
     # We could conditionally set an environment variable, but unfortunately
     # the syntax to access that is different on Windows vs Linux.
+    # Note: we also omit `--no-build` on Linux due to dotnet SDK bug 29742, so that `dotnet test` won't complain about pre-built DLLs
     - name: Test on Linux
-      run: dotnet test -f ${{ matrix.framework }} --no-build -c Release --filter TestCategory!=RequiresUI -- NUnit.TestOutputXml=TestResults
+      run: dotnet test -f ${{ matrix.framework }} -c Release --filter TestCategory!=RequiresUI -- NUnit.TestOutputXml=TestResults
       if: matrix.os == 'ubuntu-latest'
 
     - name: Test on Windows
       run: dotnet test --no-build -c Release --filter TestCategory!=SkipOnBuildServer -- NUnit.TestOutputXml=TestResults
       if: matrix.os == 'windows-latest'
 
+    # TODO: Give each test result its own name based on OS and framework and change the test-results.yml workflow accordingly, so that they don't overwrite each other
     - name: Upload Test Results
       if: always()
       uses: actions/upload-artifact@v4
--- a/.github/workflows/ci+cd.yml
+++ b/.github/workflows/ci+cd.yml
@@ -26,6 +26,11 @@ jobs:
       fail-fast: false
       matrix:
         os: [ubuntu-latest, windows-latest]
+        framework: [net461, net6.0, net8.0]
+        exclude:
+          # dotnet on Linux cannot build net461 without additional, unnecessary, work
+          - os: ubuntu-latest
+            framework: net461
     concurrency:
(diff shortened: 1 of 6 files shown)


rmunn:
Okay, now the tests are passing but the `dotnet test` run is still exiting with exit code 1 for some reason?

```plaintext
2024-10-29T11:46:47.1463558Z Passed!  - Failed:     0, Passed:   966, Skipped:    20, Total:   986, Duration: 25 m 49 s - LibChorus.Tests.dll (net6.0)
2024-10-29T11:46:47.2241280Z ##[error]Process completed with exit code 1.
```

I'll investigate why `dotnet test` doesn't like what we're doing. But it's probably related to running the tests in parallel, because the same code passes in PR #356.

rmunn:
Looks like this might be running into [dotnet/sdk#29742](https://www.github.com/dotnet/sdk/issues/29742) where running `dotnet test --no-build` after `dotnet build` is failing on ubuntu-latest runners. As a workaround, I'm going to try just running `dotnet test` without the `--no-build` parameter; this will build twice, but it should then stop complaining about DLLs being "invalid arguments" (?).

hahn-kev:
Talked about this with Robin, the issue is that Chorus.Tests only run against net461, so when we say `dotnet test -f net8.0` it fails because Chorus.Tests does not support net8.0. To fix this we will split out our `dotnet test` command to be explicit about which test project we are running, then we can skip `Chorus.Tests` when we are not running our `net461` tests. As for the tests filter which is why we split linux and windows, we can turn that into a github action expression, this will let us avoid having 6 commands,  3 per project and 2 per os.

As for the `--no-build` issue, it doesn't look like that's effecting us in our other tests, so I assume it's fine here.

rmunn:
Turns out we don't actually need the test filter to be different between Linux and Windows anymore, since we're skipping running net461 tests on Linux, and all the tests that needed the filter are net461 (and not likely to change since they use Windows Forms). So I chose to keep it simple and not have any filter differences (and just skip the SkipOnBuildServer tests everywhere).

hahn-kev:
nice work, though it looks like the Build Windows installers job is now failing.

Also it looks like no tests are actually running for net461, so I'm going to look into that

rmunn:
> nice work, though it looks like the Build Windows installers job is now failing.

Fixed in commit d5cb084.

rmunn:
@hahn-kev - Rebased on top of `master` after merging the other two PRs, so my bugfixes to Platform.IsMono are now part of this PR's base and it's no longer giving failures on Linux.

@jasonleenaylor - Please let us know if this causes any problems for the Chorus installer build.

rmunn:
There's a chicken-and-egg issue now, which can only be solved by someone with admin rights on this repo (which I don't have):

image

The `build-and-test` workflows are still expected to report their status, but this workflow removes them and replaces them with a parallel build matrix of workflows, with different names. Once this PR is merged, we'll be able to change what checks are expected, but GitHub won't allow me to merge this PR until the required status checks have passed, which includes the build-and-test checks that this PR removes.

Someone with admin access will need to check the checkbox to allow merging the PR before all the required checks have completed; I don't have admin access to the Chorus PR so I can't do it myself.

rmunn:
* Run tests in parallel, one per framework

Now that tests run on three frameworks (net461, net6.0, net8.0), we can
save a lot of build time by running each framework's tests in parallel.

* Ensure parallel test runs don't cancel each other
* ChorusMerge tests should run on all frameworks

Since ChorusMerge runs on net461;net6.0;net8.0, the ChorusMerge tests
should be run on all the same frameworks.

* Explicitly specify test project per step

Also only run Chorus.Tests on net461, since it's not defined for net6.0
and net8.0.

* lowered NUnit3TestAdapter version to a version which supports net461
* run Chorus hub tests in CI
* Checkout sources before building installer

Now that the installer job is not running in the same job matrix as the
tests, we need to re-run the dotnet restore and build steps again.

* Remove net6.0 targets

Everything that uses Chorus now runs on net8.0 or higher.

---------

Co-authored-by: Kevin Hahn <kevin_hahn@sil.org>

PR #356, opened 2024-10-29
Replace Platform.IsMono with actual OS checks

### LINKED ISSUE #355: Fix occurrences of Platform.IsMono
We no longer run Chorus tests under mono, and there are a few places where IsMono is being used as a substitute for IsLinux which are no longer correct. This results in Windows-only code such as LongToShortConverter.GetShortPathName being run on Linux when it shouldn't be.

We should go through, find all occurrences of Platform.IsMono, and update them to match the check that's actually needed. In some cases, where the check is working around a Mono-only bug, the check should stay as-is. But in other cases, we actually want to know if we're running on Windows or not, so we should really be checking Platform.IsWindows or Platform.IsUnix. Otherwise LfMerge and other code running on Linux using dotnet will be running the wrong code branch.

rmunn:
Fix #355.

Some cases of Platform.IsMono are workarounds for bugs in older versions of mono. Those can probably be removed, but for safety's sake, we'll leave them alone for now.

Other cases are clearly trying to check the platform, to run Linux-only or Windows-only code. Those cases should be replaced by Platform.IsUnix or Platform.IsWindows as appropriate, because nearly everywhere that Chorus is being run on Linux these days, it's running under dotnet rather than mono, and the Platform.IsMono check returns false.

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/chorus/356)

### DIFF
--- a/src/LibChorus/FileTypeHandlers/text/TextFileTypeHandler.cs
+++ b/src/LibChorus/FileTypeHandlers/text/TextFileTypeHandler.cs
@@ -17,6 +17,8 @@ namespace Chorus.FileTypeHandlers.text
 	[Export(typeof(IChorusFileTypeHandler))]
 	public class TextFileTypeHandler : IChorusFileTypeHandler
 	{
+		//NB: there is a post-build step in this project which copies the diff3 folder into the output on Windows
+		private static string _diff3Exe = Platform.IsWindows ? "diff3/bin/diff3.exe" : "diff3";
 		internal TextFileTypeHandler()
 		{}
 
@@ -64,7 +66,7 @@ namespace Chorus.FileTypeHandlers.text
 		{
 			//NB: surrounding with quotes didn't cut it to get past paths with spaces
 
-			return RunProcess("diff3/bin/diff3.exe", "-m " + LongToShortConverter.GetShortPath(oursPath) + " " +
+			return RunProcess(_diff3Exe, "-m " + LongToShortConverter.GetShortPath(oursPath) + " " +
 				LongToShortConverter.GetShortPath(commonPath) + " " +
 				LongToShortConverter.GetShortPath(theirPath));
 		}
@@ -81,7 +83,6 @@ namespace Chorus.FileTypeHandlers.text
 			p.StartInfo.RedirectStandardError = true;
 			p.StartInfo.RedirectStandardOutput = true;
 
-			//NB: there is a post-build step in this project which copies the diff3 folder into the output
 			p.StartInfo.FileName = filePath;
 			p.StartInfo.Arguments = arguments;
 			p.StartInfo.CreateNoWindow = true;
--- a/src/LibChorus/FileTypeHandlers/text/TextFileTypeHandler.cs
+++ b/src/LibChorus/FileTypeHandlers/text/TextFileTypeHandler.cs
@@ -154,7 +154,7 @@ namespace Chorus.FileTypeHandlers.text
 
 		public static string GetShortPath(string path)
 		{
-			if (Platform.IsMono)
+			if (!Platform.IsWindows)
 				return path;
 
 			var shortPath = new StringBuilder(255);
--- a/src/LibChorus/FileTypeHandlers/text/TextFileTypeHandler.cs
+++ b/src/LibChorus/FileTypeHandlers/text/TextFileTypeHandler.cs
@@ -79,7 +79,6 @@ namespace Chorus.FileTypeHandlers.text
 		public static string RunProcess(string filePath, string[] arguments)
 		{
 			Process p = new Process();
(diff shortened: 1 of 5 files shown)


rmunn:
@hahn-kev - This bug is "in the wild" in LfMerge. If LfMerge ever encounters a merge conflict between two .txt files, it will try to run Windows-only code and promptly error out, resulting in a Mercurial rollback because the merge conflict wasn't handled. Thankfully it's extremely unlikely that anyone will check .txt files into their FieldWorks project, let alone end up with a merge conflict in those files. However, I want to prioritize reviewing and merging this bugfix so that I can push a new LfMerge release that won't try to run Windows-only code.

src/LibChorus/FileTypeHandlers/text/TextFileTypeHandler.cs
```diff
@@ -64,26 +66,36 @@ public static string GetRawMerge(string oursPath, string commonPath, string thei
...
 		protected static string SurroundWithQuotes(string path)
 		{
 			return "\"" + path + "\"";
 		}
 
-		public static string RunProcess(string filePath, string arguments)
+		public static string RunProcess(string filePath, string[] arguments)
 		{
 			Process p = new Process();
+			ProcessStartInfo startInfo = new ProcessStartInfo();
```
hahn-kev, line 82: this startInfo isn't being used as far as I can tell

src/LibChorus/FileTypeHandlers/text/TextFileTypeHandler.cs
```diff
@@ -64,26 +66,36 @@ public static string GetRawMerge(string oursPath, string commonPath, string thei
...
+			var quotedArgs = new List<string>(arguments.Length);
+			foreach (var arg in arguments)
+			{
+				if (arg.Contains(" ")) {
+					quotedArgs.Add("\"" + arg.Replace("\"", "\\\"") + "\"");
+				} else {
+					quotedArgs.Add(arg);
+				}
+			}
+			p.StartInfo.Arguments = string.Join(" ", quotedArgs);
```
hahn-kev, line 98: I don't love the idea of changing this, was there an issue with how it worked before?

src/LibChorus/FileTypeHandlers/text/TextFileTypeHandler.cs
```diff
@@ -64,26 +66,36 @@ public static string GetRawMerge(string oursPath, string commonPath, string thei
...
+				LongToShortConverter.GetShortPath(theirPath) });
 		}
 
 		protected static string SurroundWithQuotes(string path)
 		{
 			return "\"" + path + "\"";
 		}
 
-		public static string RunProcess(string filePath, string arguments)
+		public static string RunProcess(string filePath, string[] arguments)
```
hahn-kev, line 79: this is a public API so it may be called from outside this project making this a breaking change, is it really required or can we just leave it as is.

hahn-kev (changes requested):
Looks good to me, I'd rather not make any changes to RunProcess though, see comments

src/LibChorus/FileTypeHandlers/text/TextFileTypeHandler.cs
rmunn: Yes, this code was not working, because the "test projéct" project has spaces in the filename and the quoting was incorrect. See https://github.com/sillsdev/chorus/actions/runs/11571890533/job/32210729290#step:7:4936 for an example of the errors that were happening: "ChorusMerge Error: Got error 2  diff3: extra operand '/tmp/hgmerge-ha1g3_90/one~other.txt'". That extra operand is because "test projéct", unquoted, was becoming two parameters, "test" and "projéct".

Furthermore, this failure was going *unnoticed* in the tests, because the test in question was expecting ChorusMerge to fail (due to an unresolvable merge conflict). But the diff3 code here was completely wrong, and so the ChorusMerge code these tests were *supposed* to test was never actually getting called.

src/LibChorus/FileTypeHandlers/text/TextFileTypeHandler.cs
rmunn: Yep, that was leftover from an earlier attempt to solve the bug with quoting filenames with spaces. I'll remote it.

rmunn: Done in commit 3b4f7d1b.

src/LibChorus/FileTypeHandlers/text/TextFileTypeHandler.cs
rmunn: Commit 4932740a restores RunProcess to its original state and puts the quoting logic in the method that calls it.

hahn-kev (approved):

PR #358, opened 2024-11-26
Move to net462 and update libpalaso

jasonleenaylor:
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/chorus/358)


hahn-kev (approved):
looks good to me

rmunn:
I found a couple more places where net461 would need to be adjusted to net462 and pushed commit 055beb1 to deal with them. I haven't been able to test the change to l10n/l10n.proj, so please feel free to revert that change (or any of them) if it/they turn out to be wrong.

PR #359, opened 2025-02-10
Correctly encode password in ServerSettingsModel

### LINKED ISSUE #324: Password encoding is wrong in ServerSettingsModel.LogIn
The `LogIn()` method in ServerSettingsModel currently contains the following code:

https://github.com/sillsdev/chorus/blob/271b640d6ba47b0b6d6dc2bed049d10934f7ee02/src/LibChorus/Model/ServerSettingsModel.cs#L338-L342

Notice how it's declaring the content type as "application/x-www-form-urlencoded" but not actually URL-encoding the password. This [violates the spec](https://www.w3.org/TR/html401/interact/forms.html#h-17.13.4) which says:

> Forms submitted with this content type must be encoded as follows:
> Control names and values are escaped. Space characters are replaced by `+`, and then reserved characters are escaped as described in [RFC1738, section 2.2](https://www.ietf.org/rfc/rfc1738.html#section-2.2): Non-alphanumeric characters are replaced by `%HH`, a percent sign and two hexadecimal digits representing the ASCII code of the character. Line breaks are represented as "CR LF" pairs (i.e., `%0D%0A`).

This is the root cause of issues such as https://github.com/sillsdev/languageforge-lexbox/issues/255, because it is easy to generate passwords with a password generator that contain characters such as & and + (that's not an exhaustive list), which will end up being sent incorrectly by Chorus and then rejected with a password error on the other end (when "a+b" gets URL-decoded into "a b", for example).

The long-term solution to this bug is to fix how Chorus sends passwords via HTTP POST, by URL-encoding them before writing them into the POST request body. The short-term solution will have to be telling users not to put special characters into their passwords. Some special characters make it through unscathed, due to not having any special meaning when they are URL-decoded, but it's far easier to explain "no punctuation in your password" in a choose-your-password UI than to explain "the & character, and these other characters, will fail because we messed up our code".

rmunn:
The x-www-form-urlencoded MIME type says that we're URL-encoding the password, so we need to actually URL-encode the password if we want it to be handled correctly at the other end.

Fixes #324.

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/chorus/359)

### DIFF
--- a/.github/workflows/ci+cd.yml
+++ b/.github/workflows/ci+cd.yml
@@ -11,7 +11,7 @@ on:
 jobs:
   event_file:
     name: "Event File"
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-22.04
     steps:
     - name: Upload
       uses: actions/upload-artifact@v4
@@ -25,11 +25,11 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        os: [ubuntu-latest, windows-latest]
+        os: [ubuntu-22.04, windows-latest]
         framework: [net462, net8.0]
         exclude:
           # dotnet on Linux cannot build net461 without additional, unnecessary, work
-          - os: ubuntu-latest
+          - os: ubuntu-22.04
             framework: net462
     concurrency:
       group: ${{ github.workflow }}-${{ github.ref }}-${{ matrix.os }}-${{ matrix.framework }}
@@ -55,7 +55,7 @@ jobs:
 
     - name: Install python2 for test execution
       run: sudo apt-get install python2
-      if: matrix.os == 'ubuntu-latest'
+      if: matrix.os == 'ubuntu-22.04'
 
     - name: Test Chorus
       run: dotnet test src/Chorus.Tests/Chorus.Tests.csproj -f ${{ matrix.framework }} --no-build -c Release --filter TestCategory!=SkipOnBuildServer -- NUnit.TestOutputXml=TestResults
@@ -146,7 +146,7 @@ jobs:
 
   publish-nuget:
     name: "Publish NuGet package"
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-22.04
     needs: build-installers
     if: github.event_name == 'push'
     steps:
--- a/.github/workflows/test-results.yml
+++ b/.github/workflows/test-results.yml
@@ -11,7 +11,7 @@ on:
 jobs:
   test-results:
     name: Test Results
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-22.04
     if: github.event.workflow_run.conclusion != 'skipped'
 
     permissions:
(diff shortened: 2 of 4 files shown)


imnasnainaec:
Should https://github.com/sillsdev/chorus/blob/bug/password-encoding/CHANGELOG.md#fixed be updated?

rmunn:
Testing this is proving a bit more difficult than anticipated. Because the buggy code is in ServerSettingsModel, which is only used in the "login" part of the Chorus GUI, the buggy code path is only used when logging into production Lexbox. However, when you try to set someone's password on Lexbox to something with & or + in it (there are only a few characters that trigger the bug), the frontend code doesn't allow you to do it. I can create a test user in my local dev instance of Lexbox (by temporarily disabling the frontend check that forbids those characters), but then to do S/R in FieldWorks I have to check the "Custom URL" checkbox and that won't go through the buggy codepath.

To test this, I'll need to create a test user on Lexbox, then edit their data by hand in Postgres to set their password (and salt) to the password and salt that were created for my test user in local dev. Which I'm running out of time to do today.

rmunn:
I have now tested it and proven that it works. Using the characters & or + is now safe in passwords with this bugfix. I created a test account with the password `a&b+c d` and Chorus failed to log in with that password before the bugfix, but succeeded after I applied this bugfix PR to my local copy of LibChorus.dll. (I have since changed the password on that test account, so I'm not revealing any active credentials here).

Once we fix the GHA build issues (which I suspect are because `ubuntu-latest` has just been switched to `ubuntu-24.04` which no longer offers Python 2 as an installable package), then this can be safely merged with confidence that it is tested and working.

imnasnainaec (commented):
> Reviewed 1 of 1 files at r1, 1 of 1 files at r2, 2 of 2 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/chorus/359)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @rmunn)

imnasnainaec:
> > Reviewed 1 of 1 files at r1, 1 of 1 files at r2, 2 of 2 files at r3, all commit messages.
> > _[Reviewable](https://reviewable.io/reviews/sillsdev/chorus/359)_ status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @rmunn)

I reviewed all the files, but I don't have permission in this repo to "approve".

rmunn:
Note that I changed `ubuntu-latest` to `ubuntu-22.04` in this PR so that the python2 package would be available (it was removed in ubuntu-24.04), as it's still needed for testing Mercurial 3.3 builds of Chorus. That change also needs a corresponding change to the repo's checks, which are currently waiting for a test report from "build-and-test (ubuntu-latest, net8.0)" when they should instead be waiting for "build-and-test (ubuntu-22.04, net8.0)". I don't have the repo permissions to make that particular change.

hahn-kev (approved):

rmunn:
@hahn-kev - Looks like it's not auto-merging until the required status has been reported green — and it won't be reported green, because the status is looking for `ubuntu-latest` but the test is now running on `ubuntu-22.04` instead (because 24.04 removed python2, and our CI build still tries to install a python2 package).

PR #360, opened 2025-02-19
Start tracing send recieve

hahn-kev:
the package System.Diagnostics.DiagnosticSource supports all the way back to net462 and lets use configure some distributed tracing which can then be instrumented by our applications. After tracing our sync code in Lexbox, I would like to get a little more insight into what's going on in chorus. I've trace the following:

- SyncNow which is the main entry point for S&R
- push, pull, merge which are called by SyncNow
- HgRunner to trace calls to hg
- Hg resumable calls split up by chunks (I didn't go to the request level here because that's already traced by dotnet)

Tracing is similar to logging so that when no one is listening to the activity source then no activity will be created and it'll be a noop for the code in question. In order to listen to those events with OTEL, you can do something like `service.AddOpenTelemetry().WithTracing(b => b.AddSource(LibChorusActivitySource.ActivitySourceName));`
then all the activities configured in this PR will be emit traces and logged in OTEL. I suspect other application tracing frameworks would also tap in to these traces. It could also be used for analytics.

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/chorus/360)

### DIFF
--- a/src/LibChorus/LibChorusActivitySource.cs
+++ b/src/LibChorus/LibChorusActivitySource.cs
@@ -3,6 +3,7 @@
 
 using System.Diagnostics;
 using Chorus.VcsDrivers.Mercurial;
+using JetBrains.Annotations;
 
 namespace Chorus
 {
@@ -13,6 +14,7 @@ namespace Chorus
 
 		public static void TagResumableParameters(this Activity activity, string direction, HgResumeApiParameters request)
 		{
+			if (activity is null) return;
 			activity.SetTag($"app.chorus.resumable.{direction}.chunk-size", request.ChunkSize);
 			activity.SetTag($"app.chorus.resumable.{direction}.bundle-size", request.BundleSize);
 			activity.SetTag($"app.chorus.resumable.{direction}.start-of-window", request.StartOfWindow);
--- a/src/LibChorus/LibChorusActivitySource.cs
+++ b/src/LibChorus/LibChorusActivitySource.cs
@@ -1,5 +1,5 @@
-﻿// // Copyright (c) 2025-2025 SIL International
-// // This software is licensed under the MIT License (http://opensource.org/licenses/MIT)
+﻿// Copyright (c) 2025-2025 SIL International
+// This software is licensed under the MIT License (http://opensource.org/licenses/MIT)
 
 using System.Diagnostics;
 using Chorus.VcsDrivers.Mercurial;
--- a/src/LibChorus/LibChorusActivitySource.cs
+++ b/src/LibChorus/LibChorusActivitySource.cs
@@ -1,0 +1,13 @@
+﻿// // Copyright (c) 2025-2025 SIL International
+// // This software is licensed under the MIT License (http://opensource.org/licenses/MIT)
+
+using System.Diagnostics;
+
+namespace Chorus
+{
+	public static class LibChorusActivitySource
+	{
+		public const string ActivitySourceName = "SIL.LibChorus";
+		internal static readonly ActivitySource Value = new ActivitySource(ActivitySourceName);
+	}
+}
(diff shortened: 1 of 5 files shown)


src/LibChorus/LibChorusActivitySource.cs
```diff
@@ -0,0 +1,24 @@
+﻿// // Copyright (c) 2025-2025 SIL International
+// // This software is licensed under the MIT License (http://opensource.org/licenses/MIT)
```
imnasnainaec, line 2: Is the double comment (`// //`) automatically generated by https://github.com/sillsdev/chorus/blob/master/Chorus.sln.DotSettings#L3-L5?

src/LibChorus/sync/Synchronizer.cs
```diff
@@ -245,6 +250,7 @@ public void SetIsOneOfDefaultSyncAddresses(RepositoryAddress address, bool enabl
 
 	   private void SendToOthers(HgRepository repo, List<RepositoryAddress> sourcesToTry, Dictionary<RepositoryAddress, bool> connectionAttempt)
 		{
+			using var activity = LibChorusActivitySource.Value.StartActivity();
```
imnasnainaec, line 253: What do this `activity` here in `SendToOthers` (and the one in `PullFromOthers`) do if it isn't being used anywhere with `activity?.[...]`? Do you not want any `activity?.SetStatus(ActivityStatusCode.Error,[...]` or `activity?.AddException([...]` within `ThrowIfCancelPending` or `SendToOneOther` (resp., `PullFromOneSource`)?

src/LibChorus/VcsDrivers/Mercurial/HgResumeTransport.cs
```diff
@@ -810,6 +813,8 @@ private PullStatus FinishPull(string transactionId)
 
 		private PullResponse PullOneChunk(HgResumeApiParameters request)
 		{
+			using var activity = LibChorusActivitySource.Value.StartActivity();
+			activity?.TagResumableParameters("pull", request);
```
imnasnainaec, line 817: Extra blank line after this (or remove the one from `PushOneChunk`) for consistent formatting.

src/LibChorus/LibChorusActivitySource.cs
hahn-kev: not sure, it was added by my IDE, but it probably doesn't need the double comment

src/LibChorus/sync/Synchronizer.cs
hahn-kev: it adds another activity in the trace. So when `SyncNow` is called, it will be the parent of this activity which tracks the start and end time of the activity allowing us to see what's taking time in `SyncNow`, I didn't bother to track cancellation on this activity because it'll be tracked on the parent anyway.

src/LibChorus/LibChorusActivitySource.cs
```diff
@@ -0,0 +1,24 @@
...
+using Chorus.VcsDrivers.Mercurial;
+
+namespace Chorus
+{
+	public static class LibChorusActivitySource
+	{
+		public const string ActivitySourceName = "SIL.LibChorus";
+		internal static readonly ActivitySource Value = new ActivitySource(ActivitySourceName);
+
+		public static void TagResumableParameters(this Activity activity, string direction, HgResumeApiParameters request)
```
josephmyers, line 14: This method's callers are null checked, but I wonder if it'd be safer to (also) null check `activity` here, too.

josephmyers (approved):
Code looks good to me. Just a small question/comment

imnasnainaec (commented):
> Reviewed 3 of 5 files at r1, 1 of 1 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/chorus/360)* status: 4 of 5 files reviewed, 2 unresolved discussions (waiting on @hahn-kev and @josephmyers)

PR #364, opened 2025-07-14
Update GitVersion dependency to 6.3.0

### LINKED ISSUE #363: GitVersion 5 getting old, update to GitVersion 6.3.0
I recently ran into problems running the Chorus unit test suite on Linux, as GitVersion 5.10.3 complained that it couldn't find any usable version of libssl. (It was looking for a version that was too old, and my Linux install didn't have that version installed). Updating to GitVersion 6.3.0 allowed me to run the Chorus tests on Linux, and it seems like a good idea in general to keep our dependencies up-to-date.

This will require a couple changes in GitVersion.xml: the `tag:` property has been renamed to `label:` so it's less ambiguous with Git tags, and the `prevent-increment-of-merged-branch-version` property has been split: `prevent-increment-of-merged-branch-version: false` should now become:

```yaml
prevent-increment:
  of-merged-branch: false
```

rmunn:
Fixes #363.

Requires a couple changes to GitVersion.yml since GitVersion 6 renamed some of the properties we use: `tag` becomes `label`, and `prevent-increment-of-merged-branch-version` has been split into `prevent-increment.of-merged-branch` (note the period). There's also `prevent-increment.when-current-commit-tagged` which we aren't currently using.

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/chorus/364)

### DIFF
--- a/GitVersion.yml
+++ b/GitVersion.yml
@@ -3,20 +3,21 @@ assembly-file-versioning-format: '{Major}.{Minor}.{Patch}.{env:GITVERSION_BUILD_
 mode: ContinuousDeployment
 branches:
   master:
-    tag: beta
+    label: beta
     regex: (origin/)?master
   hotfix:
-    tag: beta
+    label: beta
     regex: (origin/)?hotfix[/-]
     increment: Patch
-    prevent-increment-of-merged-branch-version: false
+    prevent-increment:
+      of-merged-branch: false
     track-merge-target: false
     tracks-release-branches: false
     is-release-branch: false
   pull-request:
     mode: ContinuousDeployment
-    tag: PR
+    label: PR
   feature:
     regex: feature[/-]
     mode: ContinuousDeployment
-    tag: alpha
+    label: alpha
--- a/l10n/l10n.proj
+++ b/l10n/l10n.proj
@@ -4,7 +4,7 @@
     <TargetFrameworks/>
     <TargetFramework>netframework4.8</TargetFramework>
     <PackageId>SIL.Chorus.l10ns</PackageId>
-    <Version>$(GitVersion_NuGetVersion)</Version>
+    <Version>$(GitVersion_SemVer)</Version>
     <Authors>Jason Naylor</Authors>
     <Company>SIL International</Company>
 	<RestartBuild Condition="'$(PkgGitVersion_GitVersion_MsBuild)' == ''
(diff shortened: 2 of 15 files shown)


ermshiperete (approved):
LGTM.

imnasnainaec:
@rmunn The installer build is failing with a missing `-Version` argument (https://github.com/sillsdev/chorus/actions/runs/16259576514/job/45905094278?pr=364#step:8:34), because two files (`l10n/l10n.proj` and `src/Installer/ChorusMergeModule.wixproj`) use `$(GitVersion_NuGetVersion)`, which appears to be less automatically available with GitVersion v6.

rmunn:
> @rmunn The installer build is failing with a missing `-Version` argument (https://github.com/sillsdev/chorus/actions/runs/16259576514/job/45905094278?pr=364#step:8:34), because two files (`l10n/l10n.proj` and `src/Installer/ChorusMergeModule.wixproj`) use `$(GitVersion_NuGetVersion)`, which appears to be less automatically available with GitVersion v6.

I see. They recommend replacing GitVersion_NuGetVersion (which would produce version strings like -beta0017) with GitVersion_SemVer, which produces version strings like -beta.17. We can reconstruct the old _NuGetVersion variable if we need to, by using `$(GitVersion_MajorMinorPatch)$(PreReleaseLabelWithDash)` and then appending a 0-padded `$(PreReleaseNumber)`.

Alternately, if we know we no longer need to support any versions of NuGet.exe older than 4.3.0 (which is QUITE old), then we can simply use `GitVersion_SemVer` and get version strings that comply with SemVer 2.0 and are more easily parseable with modern tooling.

If you think we need an exact equivalent to GitVersion_NuGetVersion then let me know, otherwise I'll update the build to use GitVersion_SemVer.

rmunn:
Incidentally, [this is the GitVersion commit](https://www.github.com/GitTools/GitVersion/commit/5b8837a471fc7dc1d6240dbb01a94f961548189b) that removed those variables.

jasonleenaylor:
I have a concern about the version continuity and the version parsing strategies that are in use in existing projects. The build system we have in place would all need to be exercised out to the tip (installation) of each affected tool. FieldWorks, OneStoryEditor, LFMerge, ??

rmunn:
Good point Jason, I've been assuming that versioning systems would handle `beta0016` being followed by `beta.17` just fine, but perhaps this is best done immediately after a new release so that there aren't any `-beta0001` that get followed by `-beta.2`. Period is U+0046 while zero is U+0048, so a naive sorting algorithm would think that `-beta0001` was a higher version number.

I'll back off the GitVersion update to the latest version of GitVersion 5.x, I believe it's 5.12.0, so that we can keep using NuGetVersion. Hopefully that will still fix https://github.com/sillsdev/chorus/issues/363 for me and allow me to build Chorus on Linux again.

PR #367, opened 2026-02-26
Exceptions during hg push should be errors

### LINKED ISSUE #366: Chorus doesn't call IProgress.WriteException on push errors
In https://github.com/sillsdev/languageforge-lexbox/issues/2057 we encountered a situation where the `hg push` operation failed due to a 500 error from the server, but everything else up to that point had gone smoothly. The result was that Chorus never called WriteException on the IProgress instance it had, so that instance still had `.ErrorEncountered = false` and the FieldWorks Lite code (which relies on ErrorEncountered) failed to inform the user that the Send/Receive had failed and the user should try again.

Looking into the Chorus code, I discovered this:

https://github.com/sillsdev/chorus/blob/1ab24b9cd13563145e98a43e1551518c8f9cc303/src/LibChorus/VcsDrivers/Mercurial/HgRepository.cs#L295-L306

The corresponding PullFromTarget code will write a progress message and then throw an exception, which eventually will result in the exception being logged with IProgress.WriteException, which will set ErrorEncountered to true. But PushToTarget doesn't throw, and doesn't call .WriteException — so FW Lite has no way of detecting this error short of brittle, error-prone things like parsing the strings Chorus writes for text like "abort: HTTP Error 500".

It would be nice if Chorus would ensure that ErrorEncountered was set to true when a push fails.

rmunn:
The IProgress implementations have an ErrorEncountered flag that is set when their WriteException or WriteError methods are called. But PushToTarget was only calling WriteMessageWithColor, so if the only exception during a Send/Receive was happening during the hg push step, then ErrorEncountered was incorrectly still false.

Fixes #366.

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/chorus/367)

### DIFF
--- a/src/LibChorusTests/sync/Synchronizer.BadSituationTests.cs
+++ b/src/LibChorusTests/sync/Synchronizer.BadSituationTests.cs
@@ -69,13 +69,65 @@ namespace LibChorus.Tests.sync
 					DoPullFromOthers = false,
 					DoSendToOthers = true,
 				};
-				options.RepositorySourcesToTry.Add(new PretendConnectableHttpRepositoryAddress());
+				options.RepositorySourcesToTry.Add(new FailingConnectableHttpRepositoryAddress("test-repo"));
 
 				var result = sender.SyncWithOptions(options);
 				Assert.That(result.Succeeded, Is.False);
 				Assert.That(result.ErrorEncountered, Is.Not.Null);
-				Assert.That(result.ErrorEncountered.Message, Does.Contain("Failed to send to"));
-				Assert.That(sender.GetProgressString(), Does.Contain("Failed to send to"));
+				Assert.That(result.ErrorEncountered.Message, Does.Contain("Failed to send changes"));
+				Assert.That(sender.GetProgressString(), Does.Contain("Failed to send to test-repo"));
+			}
+		}
+
+		[Test]
+		public void Sync_FirstPushFails_OtherSourcesStillTried()
+		{
+			using (var sender = new RepositorySetup("sender"))
+			{
+				sender.AddAndCheckinFile("one.txt", "sender's content");
+
+				var options = new SyncOptions
+				{
+					DoMergeWithOthers = false,
+					DoPullFromOthers = false,
+					DoSendToOthers = true,
+				};
+				options.RepositorySourcesToTry.Add(new FailingConnectableHttpRepositoryAddress("first-test-repo"));
+				options.RepositorySourcesToTry.Add(new FailingConnectableHttpRepositoryAddress("second-test-repo"));
+
+				var result = sender.SyncWithOptions(options);
+				Assert.That(result.Succeeded, Is.False);
+				Assert.That(result.ErrorEncountered, Is.Not.Null);
+				// Both targets should have been attempted even though the first failed.
+				var progress = sender.GetProgressString();
+				Assert.That(progress, Does.Contain("Failed to send to first-test-repo"));
+				Assert.That(progress, Does.Contain("Failed to send to second-test-repo"));
+			}
+		}
+
+		[Test]
+		public void Sync_OneOfMultiplePushTargetsFails_SyncStillSucceeds()
+		{
+			using (var sender = new RepositorySetup("sender"))
+			{
+				sender.AddAndCheckinFile("one.txt", "sender's content");
+
+				using (var receiver = new RepositorySetup("receiver", sender))
+				{
+					var options = new SyncOptions
+					{
+						DoMergeWithOthers = false,
+						DoPullFromOthers = false,
+						DoSendToOthers = true,
+					};
+					options.RepositorySourcesToTry.Add(new FailingConnectableHttpRepositoryAddress("test-repo"));
+					options.RepositorySourcesToTry.Add(receiver.GetRepositoryAddress());
+
+					var result = sender.SyncWithOptions(options);
+					Assert.That(result.Succeeded, Is.True);
+					Assert.That(result.ErrorEncountered, Is.Null);
+					Assert.That(sender.GetProgressString(), Does.Contain("Failed to send to test-repo"));
+				}
 			}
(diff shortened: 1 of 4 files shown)


rmunn:
Note that this might change the output of any code that uses SIL.Windows.Forms.Progress.LogBox as the progress dialog, because LogBox writes a full exception stacktrace in its WriteException implementation, rather than just the exception message as the existing code was doing. If this matters, a possible fix would be `WriteError(err.Message)` instead of WriteException, but that would potentially lose useful debugging info that users might copy and paste into an email. So I chose to use WriteException here, but I'll change it if requested.

myieye:
@rmunn I refactored a tad:
You recorded the exception, but the `SyncResults` object still had `Succeeded == true` and didn't know about the exception, which didn't feel quite right.
So, I mimicked exception handling in `PullFromOthers` more closely. Essentially, I just let exceptions bubble all the way up to `Synchronizer.SyncNow`. Then all 3 "error things" happens:
- Succeeded is marked as false
- The result object gets the error
- The progress object gets the error

I also added a test to demonstrate that.

I kept the one fairly specific error message "Could not update the actual files after pushing to..." by wrapping it in an `ApplicationException`, which is pretty generic, but used heavily in the `Execute` method, which the try is wrapping. So, it seemed like a fine choice.

jasonleenaylor (commented):
So there is one thing this PR does not take into account that I think does matter.
Previously if a client tried to sync to both USB and a resumeable server a failure in the remote server sync would not stop the usb sync.  Now a failed sync on one source would block the other attempts.

> @jasonleenaylor reviewed 3 files and all commit messages, and made 1 comment.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/chorus/367#-:-OrLK9dKDU8B_WBg6d37:bo7kuup)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on rmunn).

myieye:
@jasonleenaylor 

You're right, I didn't think about that.
But, it's not quite that simple. 🤔 

It is already the case that a failure during the **pull** step will fail the entire sync.
All I've done is apply that same behavior to the **push** step.

So, the only difference now is this quite precise case:
- A sync is happening to **multiple** sources
- One of those sources is a server and it's listed BEFORE other sources
- Chorus CAN connect to the server
- The PULL from the server is successful
- The PUSH to the server is not successful
- => Then pushes will not be done/attempted to sources listed AFTER it.

**That seems reasonable to me.**
But, if you think that would negatively affect users we can revert to @rmunn's initial approach.

### In summary, the options I see are:

1) Robin's initial implementation - Leave pull and push inconsistent:
- Failed pulls fail the whole sync and result in `SyncResults.Succeeded == false`
- Failed pushes do neither, but DO log exceptions to `_progress`. (this logging would be the only change to current behavior. The exception just really needs to be exposed to the caller in a clean way somehow)

(This would work fine for us and be the smallest change. It just seems semantically odd.)

2) My current implementation - Make pull and push consistent:
- Both failed pulls and pushes fail the whole sync
- They record `SyncResults.Succeeded == false` and "log" exceptions to both `_progress` and `SyncResults`.

src/LibChorusTests/sync/Synchronizer.BadSituationTests.cs
```diff
@@ -518,5 +542,27 @@ public void Sync_NewFileWithNonAsciCharacters_FileAdded()
 			}
 		}
 
+		/// <summary>
+		/// Test double that bypasses HttpRepositoryPath's remote validation.
+		/// HttpRepositoryPath.CanConnect() validates the endpoint and aborts if invalid, preventing
+		/// sync operations from executing. This allows CanConnect() to return true while sync operations
+		/// fail, enabling tests of sync failure scenarios (push/pull/merge failures).
+		/// </summary>
+		private class PretendConnectableHttpRepositoryAddress : RepositoryAddress
```
papeh: I misinterpreted this name before reading the comment. Would FalseConnectableHttpRepositoryAddress or ConnectableButUnsyncableRepositoryAddress be clearer? But it's already a long name and is commented, so also fin as is.

papeh (approved):

jasonleenaylor:
Send/Receive is both a collaboration and a backup feature for some users.
I agree it is a fairly specific edge case, but the impact of a user encountering that edge case and the type of user who might encounter it are worth considering.
Consider a user in a location with unreliable internet (the main reason that USB is still an option) trying to do Send/Receive with both resumable and usb options. They may find themselves having to disable the resumable sync in order to get it onto a USB stick as a backup or as a sneaker net collaboration. While that is an available workaround and seems fine for technical users it is a bigger hurdle for translators in the field who may not have as many computer skills.

src/LibChorusTests/sync/Synchronizer.BadSituationTests.cs
myieye: I opted for something else: `FailingConnectableHttpRepositoryAddress`.
Does that seem decent to you?

myieye:
@jasonleenaylor 

The flow is now:
- Always try sending to ALL targets
- Log all exceptions to `_progress`
- Only if ALL targets fail, throw a `SychronizationException` wrapping any/all individual push exceptions, which then marks the sync as failed

I've added two more tests that demonstrate:
1) Failing to push to one target does not stop pushing to later targets
2) Failing to push to one of multiple targets does not mark the sync as failed (I'm not sure if this is really important, but marking the whole sync as failed would be a bigger change, so I'm being conservative)

jasonleenaylor (approved):
:lgtm:

> @jasonleenaylor reviewed 2 files and all commit messages, and made 1 comment.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/chorus/367#-:-Ors_PP5Ann1Bin9Sllb:bnfp4nl)* status: all files reviewed, 1 unresolved discussion (waiting on papeh and rmunn).

PR #369, opened 2026-04-28
WIP: [API] Update URL format for Test Lexbox

papeh:
Before, production users could log in and list projects by API: https://admin.languagedepot.org/api/user/{Username}/projects and test users by
https://admin-qa.languagedepot.org/api/user/{Username}/projects Now, the 'admin' prefix is obsolete and available only in production. I assume but have not verified that production users can use https://languagedepot.org/api/user/{Username}/projects and testers
https://staging.languagedepot.org/api/user/{Username}/projects

Trim the leading '.' or '-' needed for prepending remaining subdomains in API calls.

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/chorus/369)

### DIFF
--- a/src/LibChorus/Model/ServerSettingsModel.cs
+++ b/src/LibChorus/Model/ServerSettingsModel.cs
@@ -334,7 +334,7 @@ namespace Chorus.Model
 		private WebResponse LogIn()
 		{
 			var privateQuery = IsPrivateServer ? "?private=true" : string.Empty;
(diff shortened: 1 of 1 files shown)


papeh:
This is technically a breaking change, but we don't expect it to break anything that is currently working (unless someone has a self-hosted Language Depot). Does it need a +semver:major?

PR #379, opened 2026-06-22
Fix IndexOutOfRangeException for query segments without '=' in ChorusHubServerInfo

### LINKED ISSUE #372: IndexOutOfRangeException in ChorusHubServerInfo.GetValue when query segment has no '='
In `ChorusHubServerInfo.GetValue` (`src/LibChorus/ChorusHub/ChorusHubServerInfo.cs`), the code splits each URL query segment on `=` and then accesses `parts[1]` without verifying that the split produced at least two elements:

```csharp
var parts = segment.Split('=');
if (parts.Length < 1)
    continue;
var key = parts[0].Trim(new[] { '?', ' ' });
var val = parts[1].Trim(); // throws if segment has no '='
```

The guard `parts.Length < 1` is logically unreachable — `String.Split` always returns an array with at least one element. Any query segment that contains no `=` sign (e.g., a bare flag) will therefore throw an `IndexOutOfRangeException` when `parts[1]` is accessed.

**Fix:** Change the guard to `if (parts.Length < 2) continue;`

imnasnainaec:
Two bugs in the manual query-string parser in \`ChorusHubServerInfo.GetValue\`:

1. **Unreachable guard**: \`parts.Length < 1\` was never true — \`String.Split\` always returns at least one element. A bare flag segment (no \`=\`) would throw \`IndexOutOfRangeException\` on \`parts[1]\`. Fix: change the guard to \`parts.Length < 2\`.

2. **Truncated values**: Splitting on all \`=\` characters turned \`key=val=ue\` into a 3-element array; only \`parts[1]\` was used, silently dropping everything after the second \`=\`. Fix: use \`Split(new[] { '=' }, 2)\` so the full remainder becomes \`parts[1]\`.

\`HttpUtility.ParseQueryString\` was considered but rejected: it URL-decodes values while \`ToString()\` does not URL-encode them, creating a round-trip asymmetry. Fixing that with \`Uri.EscapeDataString\` would encode colons in IPv6 addresses, breaking old clients. Since all values are machine-generated and will never contain \`+\` or \`%\`, the manual parser is correct with just these two fixes.

Fixes #372

Devin review: https://app.devin.ai/review/sillsdev/chorus/pull/379

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/chorus/379)

### DIFF
--- a/src/ChorusHubTests/ChorusHubTests.cs
+++ b/src/ChorusHubTests/ChorusHubTests.cs
@@ -247,7 +247,6 @@ namespace ChorusHubTests
 		[Test]
 		public void Parse_SegmentWithoutEquals_DoesNotThrow()
 		{
-			// A bare flag segment (no '=') previously caused IndexOutOfRangeException on parts[1].
 			const string input = "ChorusHubInfo?version=3&address=192.168.1.1&port=5912&hostname=mypc&flag";
 			Assert.DoesNotThrow(() => ChorusHubServerInfo.Parse(input));
 		}
@@ -263,8 +262,6 @@ namespace ChorusHubTests
 		[Test]
 		public void Parse_ValueContainsEquals_FullValuePreserved()
 		{
-			// Previously Split('=') on "hostname=my=pc" produced ["hostname","my","pc"];
-			// only parts[1] was used, silently dropping "=pc".
 			const string input = "ChorusHubInfo?version=3&address=192.168.1.1&port=5912&hostname=my=pc";
 			var info = ChorusHubServerInfo.Parse(input);
 			Assert.That(info.HostName, Is.EqualTo("my=pc"));
--- a/src/ChorusHubTests/ChorusHubTests.cs
+++ b/src/ChorusHubTests/ChorusHubTests.cs
@@ -266,5 +266,13 @@ namespace ChorusHubTests
 			var info = ChorusHubServerInfo.Parse(input);
 			Assert.That(info.HostName, Is.EqualTo("my=pc"));
 		}
+
+		[Test]
+		public void Parse_SegmentWithoutEquals_ParsedWithEmptyValue()
+		{
+			const string input = "ChorusHubInfo?version=3&address=192.168.1.1&port=5912&hostname";
+			var info = ChorusHubServerInfo.Parse(input);
+			Assert.That(info.HostName, Is.EqualTo(""));
+		}
 	}
 }
--- a/src/ChorusHubTests/ChorusHubTests.cs
+++ b/src/ChorusHubTests/ChorusHubTests.cs
@@ -240,4 +240,34 @@ namespace ChorusHubTests
 			}
 		}
 	}
+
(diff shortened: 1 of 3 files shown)


src/ChorusHubTests/ChorusHubTests.cs
```diff
@@ -240,4 +240,31 @@ public void GetRepostoryNames_TwoItemsInHubFolder_GetNoneForProjectFilter()
...
 	}
+
+	[TestFixture]
+	public class ChorusHubServerInfoParseTests
+	{
+		[Test]
+		public void Parse_SegmentWithoutEquals_DoesNotThrow()
+		{
+			const string input = "ChorusHubInfo?version=3&address=192.168.1.1&port=5912&hostname=mypc&flag";
+			Assert.DoesNotThrow(() => ChorusHubServerInfo.Parse(input));
```
rmunn, line 251: I feel like it would be useful to check if query parameters like `flag` have actually been parsed (since no value was given, they should probably be given the empty string as a value). Right now the code is throwing away the `flag` query parameter, so although it doesn't throw an exception any more, the query param has not been parsed and thus cannot modify ChorusHub behavior.

src/LibChorus/ChorusHub/ChorusHubServerInfo.cs
```diff
@@ -53,11 +53,10 @@ public static ChorusHubServerInfo Parse(string parameters)
...
 		{
 			var queryParameters = new NameValueCollection();
-			var querySegments = parameters.Split('&');
-			foreach (var segment in querySegments)
+			foreach (var segment in parameters.Split('&'))
 			{
-				var parts = segment.Split('=');
-				if (parts.Length < 1)
+				var parts = segment.Split(new[] { '=' }, 2);
+				if (parts.Length < 2)
```
rmunn: This will throw away the `flag` parameter used in the test. Instead, I'd recommend to keep `if (parts.Length < 1) continue;`, and instead, down where you assign `val`, set it to `parts.Length < 2 ? "" : parts[1].Trim()`.

src/ChorusHubTests/ChorusHubTests.cs
imnasnainaec: Test added.

src/LibChorus/ChorusHub/ChorusHubServerInfo.cs
imnasnainaec: Done.

imnasnainaec (commented):
> @imnasnainaec reviewed all commit messages and resolved 1 discussion.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/chorus/379)* status: 0 of 3 files reviewed, all discussions resolved.

rmunn (approved):
LGTM

rmunn:
This is perfect now: fixes the bug without discarding valueless parameters, so that `&flag` can be used in the future.

PR #380, opened 2026-06-22
Fix silent data loss when record identifier is missing in XmlMergeService

### LINKED ISSUE #373: Wrong variable in null check in XmlMergeService.MakeRecordDictionary causes silent data loss
In `XmlMergeService.MakeRecordDictionary` (`src/LibChorus/merge/xml/generic/XmlMergeService.cs`), the null check at ~line 789 tests the *parameter name* `identifierAttribute` instead of the *value* `identifier` read from the XML:

```csharp
var identifier = attrValues[identifierAttribute];
if (string.IsNullOrEmpty(identifierAttribute))  // bug: should be 'identifier'
{
    mainMergeEventListener.WarningOccurred(
        new MergeWarning(string.Format("{0}: There was no identifier for the record", pathname)));
    continue;
}
```

`identifierAttribute` is validated as non-empty earlier in the call chain, so this check never fires. Records whose identifier attribute is missing from the XML are silently added to the dictionary under a `null` key instead of triggering the warning and being skipped.

**Fix:** Change `identifierAttribute` to `identifier` in the `string.IsNullOrEmpty` call.

imnasnainaec:
In \`XmlMergeService.MakeRecordDictionary\`, the null-guard tested \`identifierAttribute\` (the parameter, always validated non-empty) instead of \`identifier\` (the value read from XML). Records whose identifier attribute was absent in the XML were silently added to the dictionary under a \`null\` key instead of triggering a warning and being skipped.

Change \`identifierAttribute\` to \`identifier\` in the \`string.IsNullOrEmpty\` call.

Fixes #373

Devin review: https://app.devin.ai/review/sillsdev/chorus/pull/380

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/chorus/380)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -43,6 +43,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 
 - Prevent S&R to Internet without full URL
 - [SIL.Chorus.LibChorus] Correctly handle & and other special characters in passwords
+- [SIL.Chorus.LibChorus] Fix null-key insertion when XML record identifier attribute is absent in XmlMergeService
 
 ## [5.1.0] - 2023-03-07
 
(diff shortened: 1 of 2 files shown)


rmunn (approved):
This is a good catch; the logic was indeed supposed to flag records that don't have the attribute, which is what the bugfix will do now.

Issue #261, opened 2021-09-03
exception handling around GetDrives()?


bobeaton:
Some of the users of Chorus in OSE get the below exception stack (green screen of death). I think in the past it was when maybe they had something plugged into a USB port... Is it possible to add exception catching if GetDrives() fails? 

Msg: Invalid class 
Class: System.Management.ManagementException
Source: System.Management
Assembly: System.Management, Version=4.0.0.0, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a
Stack:    at System.Management.ManagementException.ThrowWithExtendedInfo(ManagementStatus errorCode)
   at System.Management.ManagementObjectCollection.ManagementObjectEnumerator.MoveNext()
   at SIL.UsbDrive.Windows.UsbDriveInfoWindows.GetDrives() in C:\BuildAgent\work\a868e9f1d5fee9d4\SIL.Core.Desktop\UsbDrive\Windows\UsbDriveInfoWindows.cs:line 69
   at SIL.UsbDrive.UsbDriveInfo.GetDrives() in C:\BuildAgent\work\a868e9f1d5fee9d4\SIL.Core.Desktop\UsbDrive\UsbDriveInfo.cs:line 112
   at Chorus.UI.UsbDriveLocator.ScanForUsbDrives()
   at System.Threading.ThreadHelper.ThreadStart_Context(Object state)
   at System.Threading.ExecutionContext.RunInternal(ExecutionContext executionContext, ContextCallback callback, Object state, Boolean preserveSyncCtx)
   at System.Threading.ExecutionContext.Run(ExecutionContext executionContext, ContextCallback callback, Object state, Boolean preserveSyncCtx)
   at System.Threading.ExecutionContext.Run(ExecutionContext executionContext, ContextCallback callback, Object state)
   at System.Threading.ThreadHelper.ThreadStart()
Thread: 
Thread UI culture: en-US
Exception: System.Management.ManagementException

--Error Reporting Properties--
Version: Version 4.2.7 (apparent build date: 30-Mar-2021)
CommandLine: "C:\Program Files (x86)\SIL\OneStory Editor\StoryEditor.exe" 
CurrentDirectory: C:\Program Files (x86)\SIL\OneStory Editor
MachineName: ARIKA
OSVersion: Windows 10 
DotNetVersion: 4.0.30319.42000
WorkingSet: 25501696
UserDomainName: ARIKA
UserName: Admin
Culture: en-US
[...]

PR #383, opened 2026-06-22
Fix thread safety race in UsbDrives property

imnasnainaec:
The \`UsbDrives\` property returned the live \`List<DriveInfo>\` reference while holding the lock, but callers (e.g. \`SyncStartModel\`) enumerated the result after the lock was released. The background scan thread can concurrently call \`Clear()\`/\`AddRange()\` during that enumeration, causing \`InvalidOperationException: Collection was modified during enumeration\`.

Return \`_usbDrives.ToList()\` inside the lock so callers get a snapshot.

(This is a pre-existing issue unrelated to #261, which was already addressed in #272.)

Devin review: https://app.devin.ai/review/sillsdev/chorus/pull/383

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/chorus/383)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -43,6 +43,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 
 - Prevent S&R to Internet without full URL
 - [SIL.Chorus.LibChorus] Correctly handle & and other special characters in passwords
+- [SIL.Chorus] Fix collection-modified exception when UsbDrives property is read while background scan thread updates the list
 
 ## [5.1.0] - 2023-03-07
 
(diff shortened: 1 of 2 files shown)


rmunn (approved):
This is a decent solution to solve the immediate problem. A more effective solution would involve rewriting the class to use thread-safe collections like ConcurrentDictionary (there's no ConcurrentList, but since in this case order doesn't matter, it can be simulated by using a dictionary where the keys and values are the same). But making a copy before returning it is definitely enough for our use case: we don't need the list of USB drives to be updated "live" when someone plugs a new drive in, because if the drive doesn't show up, we can exit and reenter the dialog and then the drive will appear.

So this is fine as-is, and rewriting to use a thread-safe collection isn't truly needed.

Issue #298, opened 2022-09-22
Password Encryption/Decryption will crash LibChorus on Linux .NET6


josephmyers:
ServerSettingsModel has the methods EncryptPassword and DecryptPassword. These use System.Security.Cryptography.ProtectedData, which compiles for .NET6 but throws a runtime exception on Linux.

```
var encryptedData = ProtectedData.Protect(Encoding.Unicode.GetBytes(encryptMe),
		Encoding.Unicode.GetBytes(EntropyValue), DataProtectionScope.CurrentUser);
return Convert.ToBase64String(encryptedData);
```

No one is currently using this, but if they try it will break. We need a cross-platform password solution, or one that at least adapts to platform-specific calls.

ermshiperete:
I created a cross-platform [SIL.PasswordStore](https://github.com/sillsdev/SIL.PasswordStore) library that we could/should use ([nuget package](https://www.nuget.org/packages/SIL.PasswordStore)).  This is a wrapper around OS functionality.

This would change the way we store the password, so we would no longer put it in the config file and we no longer need EncryptPassword/DecryptPassword methods. How the password is stored and encrypted would be left to the OS.

We could consider a migration path where we read the encrypted password from the config file, decrypt it, store it with  SIL.PasswordStore, then delete it from the config file.

rmunn:
As a quick mitigation, I suggest that EncryptPassword should be changed to return null if ProtectedData.Protect throws a PlatformNotSupportedException. That will have the same effect as setting `RememberPassword = false` in the ServerSettingsModel (which is what I had to do in order to make Lexbox Send/Receive tests run on Linux hosts such as GHA runners), because when RememberPassword is false, SaveUserSettings sets Settings.Default.LanguageForgePass to null.

DecryptPassword should not be called on Linux, since EncyptPassword will have failed. However, in case DecryptPassword is somehow called anyway on an unsupported platform, it should catch PlatformNotSupportedException and return its input unmodified.

I'll submit a PR to this effect later, when I'm done with my current task.

PR #384, opened 2026-06-22
Prevent ProtectedData crash on Linux by catching PlatformNotSupportedException

imnasnainaec:
\`ServerSettingsModel.EncryptPassword\` and \`DecryptPassword\` use \`System.Security.Cryptography.ProtectedData\`, which compiles for .NET 6 but throws \`PlatformNotSupportedException\` at runtime on Linux.

Guard both methods with \`!RuntimeInformation.IsOSPlatform(OSPlatform.Windows)\` and return \`null\` early on non-Windows platforms, following the same pattern [ProtectedData itself uses](https://github.com/dotnet/dotnet/blob/f7b4c5716faaee8fb8a289aed29118cad955c45f/src/runtime/src/libraries/System.Security.Cryptography.ProtectedData/src/System/Security/Cryptography/ProtectedData.cs#L421). This avoids using exceptions for predictable control flow.

On Windows, both methods also catch runtime failures:
- \`EncryptPassword\` returns \`null\` on \`CryptographicException\` (broken DPAPI subsystem) — same effect as \`RememberPassword = false\`.
- \`DecryptPassword\` returns \`null\` on \`CryptographicException\` (different Windows user account, OS reinstall) and \`FormatException\` (invalid Base64 in the stored value) rather than the raw input. Returning the input unmodified would pass a DPAPI-encrypted base64 blob as a credential on a Windows-to-Linux settings migration, causing silent auth failures.

**Known limitation**: on non-Windows platforms, \`RememberPassword = true\` silently becomes a no-op — the preference is lost on restart with no UI feedback. Addressing that properly (disabling the checkbox or surfacing a message on unsupported platforms) requires UI changes and is left for a follow-up.

**Trade-off**: when \`EncryptPassword\` catches \`CryptographicException\` on Windows and returns \`null\`, \`SaveUserSettings\` will store \`null\` and call \`Settings.Default.Save()\`, permanently wiping a previously persisted password. Before this PR, the exception would propagate and \`Save()\` would never execute, preserving the old encrypted value. \`CryptographicException\` from \`ProtectedData.Protect\` is rare on healthy Windows systems, and the password remains available in-memory via \`PasswordForSession\` for the current session, but the user will not realize their saved password was dropped until the next restart.

Mitigates #298

Devin review: https://app.devin.ai/review/sillsdev/chorus/pull/384

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/chorus/384)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -43,7 +43,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 
 - Prevent S&R to Internet without full URL
 - [SIL.Chorus.LibChorus] Correctly handle & and other special characters in passwords
-- [SIL.Chorus.LibChorus] Fix crash in password encryption/decryption on Linux and on Windows data-protection failure
+- [SIL.Chorus.LibChorus] Fix crash in password encryption/decryption on non-Windows platforms and on Windows data-protection failure
 
 ## [5.1.0] - 2023-03-07
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -43,6 +43,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 
 - Prevent S&R to Internet without full URL
 - [SIL.Chorus.LibChorus] Correctly handle & and other special characters in passwords
+- [SIL.Chorus.LibChorus] Fix crash in password encryption/decryption on Linux and on Windows data-protection failure
 
 ## [5.1.0] - 2023-03-07
 
--- a/src/LibChorus/Model/ServerSettingsModel.cs
+++ b/src/LibChorus/Model/ServerSettingsModel.cs
@@ -374,8 +374,10 @@ namespace Chorus.Model
 			if (string.IsNullOrEmpty(encryptMe))
 				return encryptMe;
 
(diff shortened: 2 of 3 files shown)


hahn-kev:
If we know that it's going to throw then we should use `!RuntimeInformation.IsOSPlatform(OSPlatform.Windows)` to return early. This is what [ProtectedData](https://github.com/dotnet/dotnet/blob/f7b4c5716faaee8fb8a289aed29118cad955c45f/src/runtime/src/libraries/System.Security.Cryptography.ProtectedData/src/System/Security/Cryptography/ProtectedData.cs#L421) uses to check.

imnasnainaec (commented):
Good point! Done.

> @imnasnainaec made 1 comment.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/chorus/384#-:-OvulYdT0TOV70a0pj8a:b-4k0685)* status: 0 of 3 files reviewed, all discussions resolved (waiting on rmunn).

imnasnainaec:
A possible follow-up issue/pr would be to completely disable the remember-password checkbox on Linux so that it's not a silent no-op.

hahn-kev (approved):

PR #393, opened 2026-08-04
Fix ChorusMerge plugin discovery on .NET

### LINKED ISSUE #392: [claude] ChorusMerge plugin discovery uses cwd on .NET (regressed from app dir on .NET Framework)
*[Claude, autonomous]*

`ChorusFileTypeHandlerCollection.CreateWithInstalledHandlers()` locates merge handlers with:

```csharp
new DirectoryCatalog(".", "*-ChorusPlugin.dll")
```

`"."` changed meaning across the runtime port:

- **.NET Framework 4.8:** resolves to the app base dir (where the `*-ChorusPlugin.dll` sit next to `ChorusMerge`). Handlers found.
- **.NET (Core / 8):** resolves to the current working directory.

`ChorusMerge.Main` never sets the working directory, and Mercurial runs the merge tool with cwd = the repository. So on .NET the plugins (which live next to `ChorusMerge.dll`) are outside the scan: no handler is found, every file falls back to `DefaultFileTypeHandler` (keep "ours"), and real merges silently drop one side's edits.

**Fix** — scan the assembly dir, not the cwd:

```csharp
new DirectoryCatalog(AppContext.BaseDirectory, "*-ChorusPlugin.dll")
```

**Evidence:** on .NET Framework 4.8, `new DirectoryCatalog(".").FullPath` returns the app base dir and does not change when the working directory changes. On .NET 8, the FieldWorks handler is discovered only when cwd is the directory containing the `*-ChorusPlugin.dll`.

hahn-kev:
closes #392
In Lexbox when we run ChorusMerge it doesn't find it's plugins unless we override the cwd between hg calling merge and starting ChorusMerge. This fixes that bug #392 and also introduces a flag to throw when no extra plugins are found. @jasonleenaylor should this be the default? I'm worried that any non FLEx clients would not want this as the default if they don't use plugins.

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/chorus/393)

### DIFF
--- a/src/LibChorus/FileTypeHandlers/ChorusFileTypeHandlerCollection.cs
+++ b/src/LibChorus/FileTypeHandlers/ChorusFileTypeHandlerCollection.cs
@@ -17,6 +17,14 @@ namespace Chorus.FileTypeHandlers
 	public class ChorusFileTypeHandlerCollection
 	{
 		/// <summary>
+		/// When set to a truthy value (1, true, or yes),
+		/// <see cref="CreateWithInstalledHandlers"/> throws if no external plugins are found.
+		/// </summary>
+		public const string kRequirePluginsEnvVarName = "ChorusRequirePlugins";
+
+		private const string PluginSearchPattern = "*-ChorusPlugin.dll";
+
+		/// <summary>
 		/// Gets the list of handlers
 		/// </summary>
 		[ImportMany]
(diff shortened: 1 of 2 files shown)


jasonleenaylor (commented):
Even though it is retired I'd use the WeSay source as the determining factor of whether it should be the default.

> @jasonleenaylor made 1 comment.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/chorus/393#-:-OzHkntd22heRi0YkQmN:b-oqx4b)* status: 0 of 2 files reviewed, all discussions resolved.

hahn-kev:
It looks like the answer is no, WeSay does not using plugins, the lift and even WeSayConfig FileTypeHandlers are embedded in libChorus. So I'll leave the default as off.

